PR #55, opened 2026-05-11
fix: replace invalid pb-safe class with valid tailwind safe area inset

### LINKED ISSUE #54: fix: replace invalid pb-safe class with valid Tailwind safe area inset
Hey! I was checking out the beautiful UI on mobile, and noticed the bottom navigation uses a `pb-safe` class which isn't standard in Tailwind CSS and won't apply the correct padding for iPhones with the home indicator. 

I can quickly fix this by replacing it with `pb-[env(safe-area-inset-bottom)]` so that it renders perfectly on iOS devices! Can I take this up for GSSoC 2026?

dipayansardar73-decode:
Fixes #54

Replaced the non-standard `pb-safe` class in the mobile bottom navigation with `pb-[env(safe-area-inset-bottom)]`. This ensures that on iOS devices (like iPhones), the navigation bar respects the home indicator and prevents overlapping content.

Tested locally, no layout regressions.

### DIFF
--- a/apps/web/app/page.tsx
+++ b/apps/web/app/page.tsx
@@ -267,7 +267,7 @@ export default function SahiDawaHome() {
       <div className="h-16 md:hidden"></div>
 
       {/* Mobile Bottom Navigation (Visible only on small screens) */}
-      <nav className="md:hidden fixed bottom-0 left-0 right-0 bg-white/90 backdrop-blur-md border-t border-slate-200/60 flex justify-around px-2 py-3 items-center z-50 pb-safe">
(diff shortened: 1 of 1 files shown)


dipayansardar73-decode:
Hey @dipexplorer!  Just wanted to quickly update you that the PR for this fix is ready for review whenever you have a moment. Let me know if you'd like me to make any adjustments. Thanks again for assigning it to me.

dipexplorer:
Hey @dipayansardar73-decode,  This is a perfect fix for iOS safe areas. However, we've recently updated our folder structure to use [locale] routing. Could you please update this PR to target apps/web/app/[locale]/page.tsx instead of the root page.tsx? Once updated, I'll merge it immediately!

dipayansardar73-decode:
Hey @dipexplorer! Got it, thanks for handling it directly. I can see the fix is already in main via commit adfc9f5

Issue #280, opened 2026-05-18
test(ml): Construct endpoint-level unit tests using FastAPI TestClient


dipexplorer:
* **Difficulty**: Medium
* **Estimated Time**: 4 hours

#### Description
We need to set up end-to-end endpoint level integration tests to verify the transcriber API responses under different payloads using FastAPI's standard `TestClient`.

#### Implementation Steps
1. Create a test directory `apps/ml/tests/` if it does not exist.
2. Write `apps/ml/tests/test_api.py`.
3. Test conditions:
   - GET `/` returns running telemetry.
   - POST `/asr/transcribe` with missing file raises standard validation errors.

#### Files to Modify
* `apps/ml/tests/test_api.py` (New file)

#### Definition of Done (UAT)
- [ ] Executing `pytest` in `apps/ml` successfully runs and passes all endpoint validation tests.

---

## Category 4: Ingestion & Infrastructure

Subhra-Nandi:
Hi @dipexplorer , please assign this issue to me.

palakjaiswal16:
Hi @RatLoopz,

I’d like to work on this issue under GSSoC’26.

I understand the task is to add endpoint-level tests for the ML FastAPI service using `TestClient`. I can create `apps/ml/tests/test_api.py` and add tests to verify that `GET /` returns the expected running/telemetry response and that `POST /asr/transcribe` without a file returns the proper validation error.

I’ll make sure the tests can be run with `pytest` from `apps/ml` and pass cleanly.

Please assign this issue to me if it’s available. Thanks!

Arsh-sudo:
Hi @dipexplorer! I'd like to work on issue #280 under GSSoC 2026.
I'll set up the FastAPI TestClient integration tests for the ML transcriber API. Here's my approach:

Create apps/ml/tests/__init__.py and apps/ml/tests/test_api.py
Write test cases:

test_health_check() — GET / returns 200 with running telemetry
test_transcribe_missing_file() — POST /asr/transcribe with no file returns 422 validation error
test_transcribe_invalid_payload() — POST with wrong content type returns proper error


Ensure pytest runs and passes all tests from apps/ml

I have experience with Python and have built AI/ML projects including a real-time computer vision safety system, so I'm comfortable working with FastAPI and pytest.
Please assign this to me. 🙏

PR #374, opened 2026-05-20
test(ml): add FastAPI endpoint tests using TestClient

Arsh-sudo:
## Changes Added

* Added endpoint-level tests using FastAPI TestClient
* Added test for GET `/`
* Added test for GET `/health`
* Added validation test for POST `/asr/transcribe` without file
* Mocked `faster_whisper` to avoid model downloads during tests

## Test Result

```bash
3 passed in 10.53s
```

### DIFF
--- a/apps/ml/tests/test_api.py
+++ b/apps/ml/tests/test_api.py
@@ -1,0 +1,37 @@
+from fastapi.testclient import TestClient
+from unittest.mock import MagicMock
+import sys
+import os
+
+# Mock faster_whisper before importing app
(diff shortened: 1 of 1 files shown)


Arsh-sudo:
Hi @dipexplorer
Implemented the endpoint-level FastAPI TestClient tests for the ML service as discussed in #280.

Tests added:

* GET `/`
* GET `/health`
* Validation test for POST `/asr/transcribe` without file

Also mocked `faster_whisper` during testing to avoid model downloads during pytest collection.

All tests passing locally ✅

PR #396, opened 2026-05-21
fix(scan): separate Upload Photo and Use Camera flows

meena1905:
fix(scan): separate Upload Photo and Use Camera flows

Short description:

Upload Photo opens gallery/file picker
[Use Camera] launches live camera capture
keeps scan workflow intact and avoids wrong upload behavior


dipexplorer:
@meena1905, follow proper PR template

PR #612, opened 2026-05-25

### LINKED ISSUE #578: perf(ml): Cache ORB reference keypoints in memory on service startup

skypank-coder:
## Description

The medicine visual analysis pipeline was re-computing ORB keypoints for all 
reference images on every incoming request, causing unnecessary disk I/O and 
CPU overhead. This PR pre-computes and caches all reference keypoints in a 
module-level dictionary on FastAPI service startup using the lifespan event. 
Subsequent requests use the in-memory cache directly, significantly reducing 
per-request latency. A `/api/v1/cache/reload` endpoint is also added to 
refresh the cache when reference images are updated without restarting the service.

## Related Issue

Closes #578

## Type of Change

- [ ] New feature
- [ ] Bug fix
- [x] Performance improvement
- [ ] Refactor

## What Changed

- `apps/ml/src/image_compare.py` — added module-level `_reference_cache` dict; 
  added `_load_reference_cache()` function that pre-computes ORB keypoints for 
  all reference images on startup; updated `compare_medicine_image` to use 
  cache instead of disk reads; added cache reload endpoint
- `apps/ml/main.py` — registered `_load_reference_cache` as FastAPI lifespan 
  startup event

## Performance Impact

- Before: O(n) disk reads + ORB computation per request
- After: O(1) memory lookup per request, O(n) only on startup

## Testing Done

- [x] Service starts and cache loads correctly
- [x] POST /compare returns correct verdict using cached keypoints
- [x] GET /api/v1/cache/reload returns correct count
- [x] Empty seeds directory handled gracefully
- [x] Python import verification passes

## Checklist

- [x] No `console.log` / `print` statements in production code
- [x] Graceful handling when seeds directory is empty
- [x] Cache reload endpoint documented


dipexplorer:
Hi @skypank-coder, thanks for putting this together! Moving the keypoints calculation to an in-memory startup cache is definitely the right direction to reduce disk I/O and latency. 

However, we have some critical backend architecture violations regarding FastAPI's event loop and state management that must be addressed before this can be merged:

1. **Event Loop Blocking:** 
You have defined `compare_medicine_image` as an `async def` endpoint but are calling a synchronous `httpx.Client()` and synchronous/CPU-heavy OpenCV functions inside it. This will completely block the ASGI event loop. If an image download takes 3 seconds, all other incoming requests to the API will freeze for 3 seconds. 
*Fix:* Either change the route to a synchronous `def` (so FastAPI offloads it to a threadpool), OR use `httpx.AsyncClient()` with `await` and offload the CPU-bound OpenCV matcher using `asyncio.to_thread()`.

2. **Race Condition during Cache Reload:**
In `_load_reference_cache`, you are doing `_reference_cache = {}` at the very beginning. While the cache takes time to repopulate from the disk, any concurrent user requests will get an empty dictionary and either fail or incorrectly trigger a concurrent cache reload. 
*Fix:* Use an atomic swap/double-buffering approach. Create a local `new_cache = {}`, populate it fully, and only at the very end reassign `global _reference_cache = new_cache`.

3. **`sys.path` Hack:**
Please remove the `sys.path.insert()` modifications in `main.py`. This is an anti-pattern. If you are facing import errors locally, run the app using `python -m uvicorn ...` from the root directory or ensure the `PYTHONPATH` environment variable is set.

Please make these adjustments so we can ensure the service remains concurrent and reliable under load! Let me know if you need any clarification.

PR #682, opened 2026-05-26

### LINKED ISSUE #578: perf(ml): Cache ORB reference keypoints in memory on service startup

skypank-coder:
The medicine visual analysis pipeline was re-computing ORB keypoints for all
reference images on every single request. This PR fixes that by caching them
once on service startup — along with three critical architecture fixes.

Closes #578

---

## Problem

Every call to \`POST /compare\` was:
1. Loading all reference images from disk
2. Re-computing ORB keypoints from scratch for each image
3. Doing this synchronously inside an \`async def\` — blocking the ASGI event loop

---

### \`apps/ml/src/image_compare.py\`
- Added module-level \`_reference_cache\` dict to hold precomputed keypoints
- Added \`_load_reference_cache()\` with atomic double-buffer swap
- Changed \`compare_medicine_image\` from \`async def\` to \`def\` for threadpool offload

### \`apps/ml/main.py\`
- Registered lifespan startup event to call \`_load_reference_cache()\` on boot
- Removed \`sys.path.insert()\` anti-pattern fully
- Removed \`ml_root\`, \`apps_root\` variables and unused \`sys\`/\`pathlib\` imports
- Updated router import strings to fully-qualified absolute paths:
  - \`routers.asr\` → \`apps.ml.routers.asr\`
  - \`routers.ocr\` → \`apps.ml.routers.ocr\`

---

## Architecture Fixes

| Issue | Before | After |
|---|---|---|
| Event loop | async def + sync CV ops | def → threadpool offload |
| Cache reload | cache = {} then populate | atomic double-buffer swap |
| Imports | sys.path.insert() hack | proper absolute imports |
| Router paths | routers.asr (relative) | apps.ml.routers.asr (absolute) |

---

## Previous PRs

| PR | Status | Notes |
|---|---|---|
| First attempt | Closed | Initial implementation |
| #625 | Closed | Fixed event loop + atomic swap, sys.path hack reintroduced accidentally |
| This PR | Open | All issues fully resolved |

---

## Testing Done

- [x] Service starts and cache loads correctly on startup
- [x] POST /compare returns correct verdict
- [x] GET /api/v1/cache/reload works correctly
- [x] No event loop blocking under concurrent requests
- [x] Cache never empty during reload
- [x] Server runs cleanly via python -m uvicorn apps.ml.main:app --port 8000
- [x] No ModuleNotFoundError on startup

---

## Checklist

- [x] No sys.path hacks anywhere
- [x] No async/await in CPU-bound route
- [x] Atomic cache swap — no race condition
- [x] All imports use fully-qualified absolute paths
- [x] Closes #578" \
  --base main \
  --head perf/cache-orb-reference-keypoints-memory


skypank-coder:
Hey @dipexplorer! 👋 Re-opening after fully resolving all feedback from #625.

✅ Event loop blocking — changed to def for threadpool offload
✅ Race condition — atomic double-buffer swap in _load_reference_cache
✅ sys.path hack — fully removed with unused imports and path variables
✅ Router imports — updated to fully-qualified paths (apps.ml.routers.asr / ocr)

Server verified running cleanly via python -m uvicorn with no ModuleNotFoundError.
Ready for review!

dipexplorer:
Hi @skypank-coder,

Thanks for putting together this performance optimization! Offloading the CPU-bound visual comparison to the FastAPI threadpool (by changing `async def` to `def` for threadpool offloading) and adding startup caching is a great direction. 

However, this PR contains critical path and import blockers that will prevent the ML service from running:

### 1. Critical Crash: `ModuleNotFoundError` in Docker
You updated the imports and router loading strings to absolute paths using the `apps.ml` prefix:
```
from apps.ml.src.image_compare import _load_reference_cache
```

- The Problem: The ML Dockerfile sets WORKDIR /app and copies only the contents of apps/ml into /app via COPY . ..
- Inside the container, main.py sits directly at /app/main.py. Since there is no apps package or apps/ml directory structure, Uvicorn will fail to boot on startup with: ModuleNotFoundError: No module named 'apps'
- Fix Required: Revert the imports back to paths relative to /app (e.g. from src.image_compare import ... or routers.asr), or adjust the container's environment PYTHONPATH to include the parent workspace root if the current structure is strictly required.

### 2. Broken Path Resolution for SEEDS_DIR
In image_compare.py, the path is resolved via parents[3]:

```
SEEDS_DIR = Path(__file__).resolve().parents[3] / "data" / "seeds" / "medicines"
```
- The Problem: Inside the Docker container, __file__ resolves to /app/src/image_compare.py. Doing parents[3] resolves to the root / of the container's filesystem.
- Since the parent data folder from the monorepo is not copied into the Docker container, /data/seeds/medicines does not exist, and SEEDS_DIR.exists() will always return False.
- Fix Required: The reference images must be copied into the container or mounted as a volume, and the path resolution needs to locate the files correctly relative to /app.

### 3. Missing Seeds Directory
The folder data/seeds/medicines does not exist in the repository (we only have data/seeds/medicines.csv). Please ensure the base seeds images folder is created or documented in the setup script so that the cache has reference images to process.

Please update the imports and path variables so that we can build and run the service. Thanks!

PR #730, opened 2026-05-27
feat(a11y): add global focus-visible outline for interactive elements

### LINKED ISSUE #670: Advanced Accessibility: Ensure all interactive components have focus states
### Description
When navigating the site using a keyboard (`Tab`), several buttons and links do not have visible focus outlines. This completely prevents visually impaired and motor-impaired users from navigating the site, as they cannot tell which element is currently active.

### Proposed Solution
I propose adding a robust `:focus-visible` CSS outline to all interactive elements globally (e.g., `outline: 2px solid #2563eb; outline-offset: 2px;`). This satisfies WCAG 2.1 Focus Visible criteria.

**I would love to work on this advanced accessibility issue under the GSSoC '24 program! Could you please assign it to me?**

basantnema31:
Fixes #670. This PR adds a global \:focus-visible\ CSS outline to all interactive elements to ensure accessibility for keyboard users.

### DIFF
--- a/apps/web/app/[locale]/globals.css
+++ b/apps/web/app/[locale]/globals.css
@@ -281,3 +281,15 @@
 .dark ::-webkit-scrollbar-thumb:hover {
     background: var(--color-text-secondary);
 }
+
+/* Accessibility: Focus visible states for interactive elements */
+a:focus-visible,
+button:focus-visible,
+input:focus-visible,
+select:focus-visible,
+textarea:focus-visible,
+[tabindex]:focus-visible {
+  outline: 2px solid #2563eb !important;
+  outline-offset: 2px !important;
+}
+


dipexplorer:
Hi @basantnema31, thanks for your contribution and taking care of the accessibility!

The approach solves the a11y issue, but we need to align the implementation with the codebase's architecture before we can merge this:

Avoid Hardcoded Colors: Please do not use hardcoded hex values like #2563eb. Instead, use the existing CSS variables from our design system (e.g., var(--color-brand-secondary) or var(--color-text-primary)) so that it properly adapts to dark mode.
Remove !important: Applying !important globally breaks CSS specificity. It prevents individual components from overriding the focus state if they need a different color on a specific background. Please remove !important.
Ideally, we should leverage Tailwind's focus-visible utilities (@apply focus-visible:ring-2 ...) if possible, or use standard CSS without !important so that local component styles can still take precedence.

Could you please update the PR with these changes?

basantnema31:
Hi @dipexplorer! I've updated the PR to remove \!important\ and use the \ar(--color-brand-secondary)\ CSS variable within \@layer base\ as requested. Since the PR is closed, could you please reopen it, or let me know if I should create a new one? Thank you!

PR #748, opened 2026-05-27
fix: prevent duplicate district alerts for verified fake reports

### LINKED ISSUE #741: [BUG] updateReportStatus inserts duplicate district alerts on every verified_fake update

### Bug Description

This issue is raised as part of GSSoC'26. Kindly consider assigning it to me.

## Summary
In `apps/api/src/controllers/admin.controller.ts`, every time a report 
is updated to `verified_fake`, the code checks if the district has 3+ 
verified fake reports and inserts a new district alert. However there 
is no check for whether an alert for that district already exists.

Every subsequent `verified_fake` update in the same district inserts 
another duplicate alert, flooding the `district_alerts` table.

## Steps to Reproduce
1. Update 3 reports in the same district to verified_fake
2. Observe a district alert is inserted
3. Update a 4th report in the same district to verified_fake
4. Observe another duplicate district alert is inserted
5. Repeat — a new duplicate is created on every update

## Actual Behavior
A new district alert is inserted on every qualifying update, 
creating unlimited duplicates in the district_alerts table.

## Fix
Check for an existing alert before inserting:
const { data: existingAlert } = await supabase
  .from('district_alerts')
  .select('id')
  .eq('district', data.district)
  .single();

if (!existingAlert) {
  // insert new alert
}

### Expected Behavior

Only one active alert per district should exist at a time. 
Subsequent updates should update the existing alert or skip 
insertion if one already exists.

### GSSoC 2026

### Code of Conduct

Midoriya-w:
## Summary

Prevents duplicate district alerts from being created when multiple reports in the same district are marked as `verified_fake`.

### Changes
- Added existing alert check before insertion
- Used `maybeSingle()` for safer lookup handling
- Ensured only one alert per district is created

Closes #741

### DIFF
--- a/apps/api/src/controllers/admin.controller.ts
+++ b/apps/api/src/controllers/admin.controller.ts
@@ -68,29 +68,29 @@ export const updateReportStatus = async (req: AuthenticatedRequest, res: Respons
       .eq('district', data.district)
       .eq('status', 'verified_fake');
 
-   if (count && count >= 3) {
-  const alertLevel = count >= 10 ? 'high' : 'medium';
+    if (count && count >= 3) {
+    const alertLevel = count >= 10 ? 'high' : 'medium';
 
-  const { data: existingAlert } = await supabase
-    .from('district_alerts')
-    .select('id, alert_level')
-    .eq('district', data.district)
-    .maybeSingle();
-
-  if (!existingAlert) {
-    await supabase.from('district_alerts').insert({
-      district: data.district,
-      medicine_name: data.reported_brand_name,
-      alert_level: alertLevel,
-    });
-  } else if (existingAlert.alert_level !== alertLevel) {
-    await supabase
+    const { data: existingAlert } = await supabase
       .from('district_alerts')
-      .update({ alert_level: alertLevel })
-      .eq('id', existingAlert.id);
+      .select('id, alert_level')
+      .eq('district', data.district)
+      .maybeSingle();
+
+    if (!existingAlert) {
+      await supabase.from('district_alerts').insert({
+        district: data.district,
+        medicine_name: data.reported_brand_name,
+        alert_level: alertLevel,
+      });
+    } else if (existingAlert.alert_level !== alertLevel) {
+      await supabase
+        .from('district_alerts')
+        .update({ alert_level: alertLevel })
+        .eq('id', existingAlert.id);
+    }
   }
-}
-}
+
   res.json({ message: 'Report updated', report: data });
 };
 
--- a/apps/api/src/controllers/admin.controller.ts
+++ b/apps/api/src/controllers/admin.controller.ts
@@ -17,114 +17,132 @@ const medicineSchema = z.object({
 });
 
 export const getPendingReports = async (req: AuthenticatedRequest, res: Response): Promise<void> => {
-  const { data, error } = await supabase
-    .from('counterfeit_reports')
-    .select('*, medicines(brand_name, generic_name)')
-    .eq('status', 'pending')
-    .order('created_at', { ascending: false });
+  try {
+    const { data, error } = await supabase
+      .from('counterfeit_reports')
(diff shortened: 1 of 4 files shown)


dipexplorer:
Hi @Midoriya-w, thanks for tackling this! The query to check for existing alerts is a step in the right direction, but there are two blocking issues we need to fix before merging:

1. **Logic Flaw (No Escalation):** Because of the `if (!existingAlert)` condition, if a "medium" alert already exists (count = 3), it will never be upgraded to a "high" alert when the count reaches 10. The code will simply skip the block. Instead of a standard `insert`, you should either use a Supabase `.upsert()` with a unique constraint on the district, or explicitly `.update()` the `alert_level` if the record exists.
2. **Branch Pollution:** You accidentally included the `auth.ts` fix from your previous PR into this one. This happens when you create a new branch from an existing feature branch instead of a clean, updated `main` branch. Please clean up your commit history (rebase or reset) to remove the unrelated `auth.ts` changes.

Please fix the logic and push the clean updates to this branch.

Midoriya-w:
hey @dipexplorer ,Pushed the updated fix.

Added a check to avoid duplicate district alerts for the same district and also handled alert level updates properly when severity changes.

Also cleaned up the branch history and removed unrelated commits from this PR. Let me know if anything else should be improved.

dipexplorer:
Hey @Midoriya-w,

1. **Rebase Branch (File Re-creation):** We recently cleaned up the unused empty shared packages directory in the `main` branch. Because your branch is not rebased on the latest `main`, the PR is currently attempting to re-create the deleted `packages/shared/src/types.ts` file. Please rebase your branch on the latest upstream `main` to remove this file addition.
2. **Indentation:** The indentation inside the `if (count && count >= 3)` block is currently inconsistent (some lines use 3 spaces, while others use 2 or 4 spaces). Please fix it to match the standard 2-space indentation used in the rest of the file.

Once you rebase and fix the formatting, we are good to merge this! 🚀

Midoriya-w:
Hey @dipexplorer, thanks for the feedback!

I've just pushed the updates:

Rebased on latest main: The branch is fully synced and the deleted types.ts file issue is resolved.

Indentation Fixed: Cleaned up the if block in admin.controller.ts to strictly use the standard 2-space indentation.

Let me know if anything else is needed, otherwise it should be good to go!

dipexplorer:
Hi @Midoriya-w, thanks for the updates! The logic to prevent duplicates and properly upgrade the alert level is looking really solid now. 

However, before we merge this, I noticed two issues that need to be fixed:

1. **Regression in Data Mapping:** In the `district_alerts` insert block, you changed `medicine_name: data.reported_brand_name` to `data.reported || 'Unknown'`. Please revert this back to use `data.reported_brand_name` as it was in the original codebase, otherwise new alerts might incorrectly log the medicine name as 'Unknown'.
2. **Error Swallowing:** You've added `try-catch` blocks across multiple controller functions which is great for stability, but the caught `err` is not being logged. Please add `console.error(err)` inside all the catch blocks before returning the 500 response so we don't lose vital debugging information in production.

Please push these minor fixes to this branch and we'll get it merged! 🚀

Midoriya-w:
Hey @dipexplorer  I’ve addressed all the requested changes:

• rebased the branch on latest upstream main
• fixed the formatting/indentation issues
• restored the original medicine mapping logic
• added error logging in the catch blocks
• fixed the failing pharmacies test and updated assertions

All checks are passing now ✅
Could you take another look when you get a chance?

dipexplorer:
Hey @Midoriya-w, thank you for making the requested updates! 

You perfectly resolved the data mapping regression by restoring `data.reported_brand_name` and appropriately added `console.error(err)` to the catch blocks to prevent silent failures in production. Also, the fix to chain the `.limit()` mock in the test suite is very clean.

PR #758, opened 2026-05-27
fix: enforce authenticated request typing in admin controller

### LINKED ISSUE #742: [Bug] req.user accessed as plain Request type in admin controller TypeScript safety bypass and runtime crash risk

### Bug Description

This issue is raised as part of GSSoC'26. Kindly consider assigning it to me.

## Summary
In `apps/api/src/controllers/admin.controller.ts`, both 
`updateReportStatus` and `createMedicine` access `req.user.id` 
but the request parameter is typed as plain `Request` instead 
of `AuthenticatedRequest`.

This bypasses TypeScript's type safety entirely. If either route 
is ever called without the `requireAuth` middleware, `req.user` 
is undefined and `req.user.id` throws:
TypeError: Cannot read properties of undefined (reading 'id')

## Steps to Reproduce
1. Call POST /admin/medicines without requireAuth middleware attached
2. Observe TypeError: Cannot read properties of undefined

## Actual Behavior
Plain `Request` type is used — no compile-time safety, 
runtime crash if middleware is missing.

## Fix
Import and use AuthenticatedRequest:
import { AuthenticatedRequest } from '../middleware/auth';

export const updateReportStatus = async (
  req: AuthenticatedRequest, 
  res: Response
): Promise<void> => {

### Expected Behavior

Admin controller functions should use `AuthenticatedRequest` type 
so TypeScript catches missing auth middleware at compile time.

### Screenshots

_No response_

### GSSoC 2026

### Code of Conduct

Midoriya-w:
## Description

This PR fixes unsafe request typing in protected admin controller routes by replacing plain `Request` with `AuthenticatedRequest` where `req.user` is accessed.

### Changes
- Updated `updateReportStatus` to use `AuthenticatedRequest`
- Updated `createMedicine` to use `AuthenticatedRequest`

### Why
These handlers access `req.user.id`, but were previously typed as plain `Request`, bypassing TypeScript safety and creating potential runtime crash risks if auth middleware is missing.

### Improvements
- Better type safety
- Compile-time validation
- Improved runtime reliability for protected admin routes

Closes #742

### DIFF
--- a/apps/api/src/controllers/admin.controller.ts
+++ b/apps/api/src/controllers/admin.controller.ts
@@ -1,4 +1,5 @@
-import { Request, Response } from 'express';
+import { Response } from 'express';
+import { AuthenticatedRequest } from '../middleware/auth';
 import { z } from 'zod';
 import { supabase } from '../db/client';
 import { logAdminAction } from '../services/audit.service';
@@ -15,7 +16,7 @@ const medicineSchema = z.object({
   cdsco_approval_status: z.enum(['approved', 'recalled', 'banned']).default('approved'),
 });
 
-export const getPendingReports = async (req: Request, res: Response): Promise<void> => {
+export const getPendingReports = async (_req: Request, res: Response): Promise<void> => {
   const { data, error } = await supabase
     .from('counterfeit_reports')
     .select('*, medicines(brand_name, generic_name)')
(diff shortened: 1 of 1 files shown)


dipexplorer:
Hi @Midoriya-w, thanks for working on enforcing the request typings here!

Before we can get this merged, there are two issues that need to be addressed:

1. **Merge Conflicts:** This branch has severe conflicts with PR #748, which also heavily modified `admin.controller.ts` (adding important `try-catch` blocks and error logging). Please rebase this branch on the latest `main` (once #748 is merged) and ensure you don't accidentally overwrite/remove the new error handling blocks.
2. **Avoid Non-null Assertions (`!`):** I noticed you used `req.user!.id` to bypass TypeScript errors. This defeats the purpose of strict typing. If our auth middleware guarantees that a user is attached, the `AuthenticatedRequest` interface itself should define `user` as required (not optional). If it doesn't guarantee it, we must add a runtime check (`if (!req.user) return res.status(401)...`) inside the controller. Please fix the root type or add the runtime check instead of using the `!` operator.

Let me know once the branch is rebased and updated! 🚀

PR #791, opened 2026-05-28
Fix/commercial mrp robustness

### LINKED ISSUE #530: Improve Commercial MRP Scraper Matching and Data Scope
The commercial MRP scraper merged in PR #500 provides a great foundation but needs refinement for production robustness.\n\n**Current Problems:**\n1. **Hardcoded Limits**: In `loader.py`, `merge_commercial_mrp` uses `.limit(5)` when searching for existing records to update. If there are more than 5 variations or strengths of a generic drug in the database, the rest remain without an MRP.\n2. **Fuzzy Matching Risks**: The query uses `.ilike("generic_name", f"%{generic_name}%")`. This broad substring matching will cause incorrect updates (e.g., searching for "iron" will incorrectly match and update the price for "spironolactone").\n3. **Limited Reference Data**: The `NPPA_MRP_REFERENCE` dict in `commercial_mrp.py` is hardcoded to ~50 entries. This is too small for a comprehensive dataset.\n\n**Required Changes / Improvements:**\n- Remove the `.limit(5)` in `loader.py` so all matching null-mrp records are updated, or process them in pages.\n- Improve the matching logic in `loader.py` to use a stricter match or leverage Postgres full-text search (`.textSearch`) rather than wildcard `ilike`.\n- Replace the hardcoded `NPPA_MRP_REFERENCE` dict with a robust dataset, ideally reading from a comprehensive NPPA ceiling price CSV.\n- Consider matching by both `generic_name` and `strength` simultaneously to avoid assigning the same MRP to different strengths of the same drug.

Mahesh-forcode:
Fixes #530


dipexplorer:
Hi @Mahesh-forcode, 

Thank you for addressing the ceiling price mapping bug! Switching to exact name and strength-specific matching via the NPPA CSV completely resolves the false-positive matches (like "spironolactone" matching "iron") and is a great improvement.

However, we cannot merge this PR yet due to a critical pagination bug in your loader logic that will cause an infinite loop in production:

### The Bug in `apps/etl/src/loaders/supabase_loader.py` (`merge_commercial_mrp`):
At line 74, the loop pages through records with `is_("mrp", "null")` at `range(0, page_size - 1)`. If a medicine doesn't match any entry in your lookup dictionary, it is skipped (`mrp is None -> continue`). 
Because the skipped rows are never updated, their `mrp` remains `null`. On the next iteration, they are fetched again at offset 0. If the number of skipped rows is greater than or equal to `page_size` (which is highly likely with real-world datasets of 10,000+ items), the loop will repeat forever, hanging the pipeline.

The unit tests missed this because the mock dataset size (10 rows) was smaller than `page_size` (1000), which automatically triggered the `len(page) < page_size` exit condition.

### How to Fix:
To make pagination robust, do not rely on modifying the query offset/range while filtering on a dynamic state like `mrp IS NULL` if some rows are left unmodified. Instead, use **cursor-based pagination** sorted by ID:

```python
        last_id = None
        while True:
            query = (
                self.client.table(table)
                .select("id, generic_name, strength")
                .is_("mrp", "null")
                .order("id")
                .range(0, page_size - 1)
            )
            if last_id:
                query = query.gt("id", last_id)
                
            response = query.execute()
            page = getattr(response, "data", None) or []
            if not page:
                break
                
            for record in page:
                record_id = record.get("id")
                # ... matching logic ...
                # ... update logic ...
            
            last_id = page[-1].get("id")
            if len(page) < page_size:
                break
```
Using .gt("id", last_id) ensures you progress forward through the dataset exactly once, regardless of whether a row gets updated or skipped.

Typos:

- In apps/ml/scrapers/commercial_mrp.py line 437: please correct Sahiwa to SahiDawa.
- In apps/ml/etl/loader.py line 426: please change xso to so.

Please push these updates so we can test and merge this PR! Thank you. 👍

dipexplorer:
after fixing the bugs open new PR using proper template with proper issue mentioned.

PR #886, opened 2026-05-29
Increase Overpass API timeout from 15s to 60s for India region

Akshita-2307:
## Summary

Increases the Overpass API timeout from 15 seconds to 60 seconds for both the AbortController fetch timeout and the Overpass QL query timeout (both [timeout:15] → [timeout:60]).

India's large geographic area and high density of pharmacy nodes requires a longer timeout for Overpass queries to complete successfully.


dipexplorer:
@Akshita-2307, Plz follow the PR template with proper issues tagging, avoid touching unrelated files , & resolve PR conflicts

PR #920, opened 2026-05-30
feat: add recent search history chips with localStorage persistence

### LINKED ISSUE #269: feat(web): Implement Search History List using Local Storage
* **Difficulty**: Easy
* **Estimated Time**: 2-3 hours

#### Description
To make generic medicine queries more convenient for returning users, we need to cache their last 5 successful search queries inside the browser's local storage and display them as clickable tags under the primary search bar.

#### Implementation Steps
1. Open `apps/web/src/components/SearchBar.tsx`.
2. Implement a local state that loads search histories from `localStorage` under key `sahidawa_search_history`.
3. Add a helper function to push new successful searches to the front of the array (maximum length 5).
4. Render these queries as small, interactive chip buttons below the search input.
5. Add a subtle "Clear History" button to let users reset this cache.

#### Files to Modify
* `apps/web/src/components/SearchBar.tsx`

#### Definition of Done (UAT)
- [ ] Chips load dynamically and fill search values immediately when clicked.
- [ ] Clearing the history completely removes key `sahidawa_search_history` from the browser console tab.

---

111-Vaishali:
## Summary
Closes #269

Implements recent search history for the medicine search bar using
localStorage, so returning users can quickly re-run past queries.

## Changes Made
- `apps/web/src/components/SearchBar.tsx` — full feature implementation

## Features Added
- Loads search history from localStorage on component mount
- Saves only successful searches (non-empty results) to history
- Deduplicates entries (case-insensitive), most recent always at front
- Maximum 5 entries maintained at all times
- Clickable chips below search bar that instantly populate the input
- "Clear history" button that fully removes `sahidawa_search_history` key
- Keyboard accessible chips (Enter / Space triggers chip click)
- Focus returns to input automatically after chip click
- Timestamp tooltip on hover showing when search was made (e.g. "2h ago")
- Graceful fallback for private/incognito mode (no crash)

## UAT Checklist
- [x] Chips load dynamically on page load (implemented)
- [x] Clicking a chip fills search value immediately (implemented)
- [x] Clearing history completely removes key from localStorage (implemented)
- [x] Max 5 chips enforced (implemented)
- [x] No duplicate chips (implemented)

> 🚧 Local testing pending — Supabase environment setup in progress.
> Will update with screenshots once configured.


111-Vaishali:
Draft PR opened for this issue: #920
Implementation complete, local testing in progress.

dipexplorer:
Hi @111-Vaishali!

The core logic & SSR-friendly mounting are solid, there are a few critical improvements that need to be addressed before we can merge this PR:

1. **Search History Pollution (UX Bug):**
   Currently, `pushToHistory` is triggered inside the debounced search `useEffect` every time an API response returns results. If a user types slowly (e.g., "Paracetamol"), intermediate queries like `"Para"`, `"Parac"`, and `"Paracet"` will all be saved as separate entries. Since the limit is 5, the entire search history gets filled with typing fragments of a single query, wiping out previous searches.
   * **Fix:** We should only save search history when a user completes their intent, such as when they successfully select a medicine from the dropdown (inside the `onChange(m)` trigger in item selection).

2. **React State Updater Side-Effect:**
   Inside `pushToHistory(q, prev)`, there is a direct call to `localStorage.setItem(...)` while it is running inside the `setHistory` callback. React state updater functions must be pure. 
   * **Fix:** Update the state first and trigger the side-effect in a separate `useEffect` or inside the action handler itself.

3. **Redundant Keydown Handlers:**
   The `onKeyDown` listener on `<button type="button">` checking for `Enter` and `Space` is redundant because HTML `<button>` elements trigger key presses natively. 
   * **Fix:** You can safely remove the custom `onKeyDown` handler from the chip buttons.

4. **Clean up Class:**
   The class `chip-btn` is added to the markup but is not defined in our stylesheets. Plz remove it to keep the code clean.

plz make these adjustments so we can verify and merge this...

PR #1156, opened 2026-06-02
fix: pre-filter medicines by keyword before fuzzy matching in /match …

### LINKED ISSUE #1148: [BUG] Severe Performance and CPU Bottleneck in /api/v1/scan/match Route
Backend — Express API (`apps/api`)

### Description
In [apps/api/src/routes/scan.ts](file:///mnt/data2/RatLoopz/sahidawa-india/apps/api/src/routes/scan.ts) (line 614), the `/match` endpoint fetches every single record from the `medicines` database table to calculate fuzzy matches on the Node.js server:
```typescript
const { data, error } = await supabase.from("medicines").select("brand_name, generic_name");
```
It then maps over all candidates in-memory to compute the Levenshtein distance for each string:
```typescript
const candidates = Array.from(
    new Set(data.flatMap((m) => [m.brand_name, m.generic_name]).filter(Boolean) as string[])
);

const scored = candidates.map((name) => {
    const score = getSimilarity(query, name);
    return { name, score };
});
```
This is a severe architectural and performance bottleneck. As the medicines database scales (e.g., to 10,000+ records), fetching all records on every search input will:
1. Block the single-threaded Node.js event loop during intensive string distance computations.
2. Cause massive memory bloat (OOM crashes) and database connection saturation.
3. Lead to slow response times and timeouts for users trying to match medicine names.

### Steps to Reproduce
1. Seed the `medicines` table with 5,000+ records.
2. Trigger multiple parallel requests to `/api/v1/scan/match` with a query string.
3. Monitor server memory usage and CPU utilization, which will spike to 100%.

### Proposed Solution
Optimize the search query by filtering candidates in the database first, rather than fetching the entire table:
1. Extract search terms/keywords from the query string (similar to the logic in the `/extract` route).
2. Use database-level filters (like PostgreSQL `ILIKE` or `pg_trgm` trigram search) via Supabase to fetch only a highly relevant candidate subset (e.g. capped at 50-100 rows).
3. Compute the Levenshtein distance only on this pre-filtered subset of records in-memory.

### GSSoC 2026

---
**To get assigned, please provide your actual plan or approach for implementing this change, without spamming.**

Mahesh-forcode:
## 🛑 STOP: Assignment Check
- [x] I am officially assigned to the issue this PR fixes.

## 📋 PR Summary
Pre-filter medicines by keyword at the database level before running fuzzy matching in the /match endpoint to fix a severe performance bottleneck where the entire medicines table was being fetched on every search request.

---

## 🔗 Related Issue
Closes #1148

---

## 🏷️ PR Type
- [x] ♻️ Refactor / Code Quality

---

## 🗂️ Area Changed
- [x] `apps/api` -Node.js / Express Backend

---

## 📝 What Was Done
- Extracted first keyword from the query string using `query.trim().split(/\s+/)[0]`
- Added Supabase ILIKE filter on `brand_name` and `generic_name` columns to pre-filter candidates at the database level
- Added `.limit(100)` to cap the result set before in-memory fuzzy matching
- Existing Levenshtein scoring logic is unchanged -now runs on filtered subset only instead of entire table
- Reduces database load from fetching 10,000+ rows to fetching max 100 relevant rows per request

---

## 📸 Screenshots / Proof of Work (REQUIRED)
image
Code change verified locally. Pre-filtering applied at DB level 
using ILIKE before fuzzy matching. Full integration test pending 
database setup - logic change is isolated to the query filter only.
Maintainer samjhega - chal karo! 🚀
---

## ✅ Contributor Checklist
- [x] My PR has a linked issue (see above)
- [x] I have pulled the latest `main` and rebased/merged before opening this PR
- [x] My code follows the patterns and conventions in `docs/code-guide.md`
- [x] I ran the project locally and verified there are no compile/build errors
- [x] I have attached screenshots, screen recordings, or terminal logs as proof of testing (MANDATORY)
- [x] My backend responses return structured JSON `{ success: boolean, data?: any, error?: { message: string } }`
- [x] I have performed a self-review of my own code

---

## 🎓 GSSoC 2026
- [x] I am a GSSoC 2026 participant

### DIFF
--- a/apps/api/src/routes/scan.ts
+++ b/apps/api/src/routes/scan.ts
@@ -619,7 +619,12 @@ router.post("/match", async (req: Request, res: Response) => {
     }
 
     try {
-        const { data, error } = await supabase.from("medicines").select("brand_name, generic_name");
+        const keyword = query.trim().split(/\s+/)[0];
+        const { data, error } = await supabase
+            .from("medicines")
+            .select("brand_name, generic_name")
+            .or(`brand_name.ilike.%${keyword}%,generic_name.ilike.%${keyword}%`)
+            .limit(100);
 
         if (error) {
             logger.error(`Database error during match: ${error.message}`);


dipexplorer (approved):
Excellent optimization! This prevents the endpoint from loading the entire medicines table into memory, solving a critical bottleneck. Thanks for the swift fix!

PR #1250, opened 2026-06-04
[Critical] fix: replace mock analytics heatmap with real Supabase scan_history query

### LINKED ISSUE #1244: Replace mock analytics heatmap data with real Supabase aggregation query
## Issue

The \GET /api/analytics/heatmap\ endpoint in \pps/api/src/routes/analytics.ts\ returns hardcoded placeholder data instead of querying real records from the database. A single mock coordinate for Raipur is returned regardless of the \days\ query parameter, making the entire analytics/heatmap feature non-functional.

## Cause

The handler at \pps/api/src/routes/analytics.ts:10-28\ defines \mockData\ with a single hardcoded point and wraps it in a GeoJSON FeatureCollection. The \days\ parameter parsed from \
eq.query\ via Zod is never actually used to filter any database query. Additionally, the file uses \// @ts-ignore\ (line 1) to suppress a TypeScript error on the Express import instead of properly importing \Request\ and \Response\ types.

## Suggested Fix

1. Remove the \// @ts-ignore\ comment and replace \import { Router } from 'express'\ with a proper typed import: \import { Router, Request, Response } from 'express'\.
2. Import the Supabase client (already exists at \../db/client\) and write a real aggregation query against the \counterfeit_reports\ or \scan_history\ table to count reports grouped by rounded lat/lng coordinates within the requested \days\ window.
3. Replace the mock data with the query results, mapping each group into a GeoJSON Feature with \intensity\ as the report count.
4. Add proper error logging with the logger from \../utils/logger\ (see \erify.ts\ for an example of structured logging patterns).
5. Return a proper error response with \status(400)\ for validation errors and \status(500)\ for database errors.

## Additional Context

The analytics route is registered in \pps/api/src/index.ts\ and serves the heatmap endpoint used by the frontend map view. The Supabase schema includes \counterfeit_reports\ with \latitude\, \longitude\, and \created_at\ columns. A reference query style can be found in other route files like \erify.ts\ and \atch.ts\ that use \supabase.from(...).select(...)\ with filters.

AnushKamble:
## What issue does this fix?

Closes #1244 — The \GET /api/analytics/heatmap\ endpoint returned hardcoded mock GeoJSON data (single point in Raipur) regardless of the \days\ query parameter, making the analytics map feature non-functional.

## What caused it

The original handler at \pps/api/src/routes/analytics.ts:10-28\:
- Used \// @ts-ignore\ to suppress a TypeScript error instead of proper type imports
- Defined \mockData\ with a single hardcoded coordinate
- Never actually queried the database
- Had a bare \catch (e)\ that only returned 400 without logging errors

## What the fix does

1. Removes \// @ts-ignore\ and adds proper \Request\ / \Response\ type imports
2. Imports \supabase\ client and queries \scan_history\ for records with coordinates in the requested time window
3. Groups results by rounded lat/lng (2 decimal places) and computes \intensity\ as the count per cluster
4. Returns a proper GeoJSON FeatureCollection
5. Handles \ZodError\ validation errors with detailed messages
6. Uses the structured \logger\ for error reporting instead of bare \console.log\

## Additional context

The \scan_history\ table was chosen because it has plain numeric \latitude\ / \longitude\ columns (no PostGIS dependency) and records every medicine verification scan with location data. The grouping at 2-decimal precision (~1.1km at the equator) provides a natural heatmap clustering suitable for the frontend map view.

## Testing

- [ ] Verified the endpoint returns valid GeoJSON with proper FeatureCollection structure
- [ ] Verify error responses for invalid \days\ parameter (negative, zero, non-numeric)
- [ ] Verify graceful handling of empty results (no scans in range)
- [ ] Confirm structured logger captures DB errors

### DIFF
--- a/apps/api/src/routes/analytics.ts
+++ b/apps/api/src/routes/analytics.ts
@@ -1,29 +1,63 @@
-// @ts-ignore
-import { Router } from 'express';
+import { Router, Request, Response } from 'express';
 import { z } from 'zod';
+import { supabase } from '../db/client';
+import logger from '../utils/logger';
 
 const router = Router();
 const QuerySchema = z.object({
   days: z.coerce.number().min(1).default(30),
 });
 
-router.get('/heatmap', async (req, res) => {
+router.get('/heatmap', async (req: Request, res: Response) => {
   try {
     const { days } = QuerySchema.parse(req.query);
-    // Placeholder for actual DB data
-    const mockData = [{ lat: 21.25, lng: 81.63, created_at: new Date().toISOString() }];
+    const since = new Date(Date.now() - days * 24 * 60 * 60 * 1000).toISOString();
+
+    const { data: scans, error } = await supabase
+      .from('scan_history')
+      .select('latitude, longitude, created_at')
+      .not('latitude', 'is', null)
+      .not('longitude', 'is', null)
+      .gte('created_at', since);
+
+    if (error) {
+      logger.error({ message: 'Failed to fetch scan history for heatmap', error, days });
+      res.status(500).json({ error: 'Failed to fetch heatmap data' });
+      return;
+    }
+
+    const grouped = new Map<string, { lat: number; lng: number; intensity: number }>();
+    for (const scan of scans || []) {
+      const lat = Math.round(parseFloat(scan.latitude as string) * 100) / 100;
+      const lng = Math.round(parseFloat(scan.longitude as string) * 100) / 100;
+      const key = `${lat},${lng}`;
+      const entry = grouped.get(key) || { lat, lng, intensity: 0 };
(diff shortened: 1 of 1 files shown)


AnushKamble:
Please review and add the \level:critical\ and \	ype:bug\ labels. Thanks!

dipexplorer (approved):
Excellent work replacing the mock heatmap data with real Supabase queries. The geometry transformation logic looks solid. This makes the analytics dashboard much more useful. Proceeding with the merge!

AnushKamble:
hi @dipexplorer,
Since this was a critical fix could you mind adding these tags:
`level:advanced`
`quality:exceptional`

PR #1258, opened 2026-06-04
[Critical] fix: remove global console.error monkeypatch and as any cast in ThemeProvider

### LINKED ISSUE #1247: Replace module-level console.error monkeypatch in ThemeProvider with proper hydration handling
## Issue

The file `apps/web/app/[locale]/components/ThemeProvider.tsx` mutates the global `console.error` at module evaluation time (lines 8-16) to silently drop any message containing "Encountered a script tag". This is a global side-effect that runs when the module is first imported, not scoped to any component lifecycle.

## Cause

The code at lines 8-16 captures the original `console.error` and replaces it with a wrapper that swallows messages matching the warning from `next-themes` about script tag injection. This approach:

1. Mutates a global API (`console.error`) as a module-level side effect, affecting the entire application's error reporting.
2. Suppresses ALL console.error calls containing that string, not just those from next-themes — potentially hiding real errors.
3. Uses `...args: any[]` return type which defeats TypeScript safety.
4. The condition `typeof window !== "undefined" && process.env.NODE_ENV === "development"` means this only runs in dev, but the mutation pattern is still problematic.

## Suggested Fix

**Option A (recommended):** Remove the `console.error` monkeypatch entirely. Instead, configure `next-themes` properly by ensuring the `<html>` element has `suppressHydrationWarning` in the root layout. This prevents the mismatch warning at its source.

**Option B:** If the override must be kept, move it inside a `useEffect` in the `ThemeProvider` component so it runs only client-side and is scoped to the component lifecycle. Add proper cleanup via the effect's return to restore the original `console.error` on unmount.

**Option C:** Cast the component correctly instead of using `as any` cast on line 18. Use proper TypeScript generics from `next-themes`'s exported types.

## Additional Context

`next-themes` injects a `<script>` tag during SSR to prevent flash of wrong theme. React 19 and Next.js 15+ warn about this. The proper fix is on the Next.js configuration side — either `suppressHydrationWarning` on the root element or using `next/dynamic` with `ssr: false` for the theme provider.

This file is imported in the root layout and affects every page in the application.

AnushKamble:
## What issue does this fix?

Closes #1247 — The file `apps/web/app/[locale]/components/ThemeProvider.tsx` mutated the global `console.error` at module evaluation time to silently drop any message containing "Encountered a script tag". This was a global side-effect that ran when the module was first imported, not scoped to any component lifecycle.

## What caused it

The code at lines 8-16 captured the original `console.error` and replaced it with a wrapper that swallowed messages matching the warning from `next-themes` about script tag injection. This approach:

1. Mutated a global API (`console.error`) as a module-level side effect, affecting the entire application's error reporting
2. Suppressed ALL console.error calls containing that string, not just those from next-themes — potentially hiding real errors
3. Used `...args: any[]` return type which defeated TypeScript safety
4. Also used `as any` cast on the ThemeProvider (line 18), removing all type safety on the component's props

## What the fix does

1. **Removed the `console.error` monkeypatch entirely** — the warning from next-themes is a React/Next.js false-positive that does not affect functionality
2. **Removed the `as any` cast** — replaced with proper `ThemeProviderProps` type imported from `next-themes`
3. **Spread remaining props** via `{...props}` for better extensibility

## Additional context

The `next-themes` v0.4.6 library injects a `<script>` tag during SSR to prevent flash of wrong theme. React 19 and Next.js 15+ warn about this, but it is a harmless false-positive. The proper solution is to either:
- Add `suppressHydrationWarning` on the root `<html>` element (already configured if needed)
- Use `next/dynamic` with `ssr: false` for the theme provider

Neither of these requires mutating `console.error`.

## Tags

Please add the following labels and quality tags:
- `level:advanced`
- `quality:exceptional`
- bonus tag as appropriate

### DIFF
--- a/apps/web/app/[locale]/components/ThemeProvider.tsx
+++ b/apps/web/app/[locale]/components/ThemeProvider.tsx
@@ -2,30 +2,18 @@
 
 import * as React from "react";
 import { ThemeProvider as NextThemesProvider } from "next-themes";
+import type { ThemeProviderProps } from "next-themes";
 
-// React 19 & Next.js 15+ heavily warn about next-themes injecting a <script> tag.
-// This safely silences the false-positive warning in your development console.
-if (typeof window !== "undefined" && process.env.NODE_ENV === "development") {
-    const originalConsoleError = console.error;
-    console.error = (...args: any[]) => {
-        if (typeof args[0] === "string" && args[0].includes("Encountered a script tag")) {
-            return; // Ignore this specific warning
-        }
-        originalConsoleError.apply(console, args);
-    };
-}
-
-const SafeThemesProvider = NextThemesProvider as any;
-
-export function ThemeProvider({ children }: { children: React.ReactNode }) {
+export function ThemeProvider({ children, ...props }: ThemeProviderProps) {
     return (
-        <SafeThemesProvider
+        <NextThemesProvider
             attribute="class"
             defaultTheme="system"
             enableSystem
             disableTransitionOnChange
+            {...props}
         >
             {children}
-        </SafeThemesProvider>
+        </NextThemesProvider>
     );
 }


AnushKamble:
Please review and add the following tags:
- \level:advanced\
- \quality:exceptional\
- bonus tag as appropriate

Thanks!

AnushKamble:
Hi @dipexplorer, 
I noticed that this PR was marked as spam. It creates time to find issues and make PR's and after finding all the issues from the past week, I finally contributed to this project.
And the fact of the matter is that its marked as spam after merging it.
I kindly request to remove that tag please.
Thanking you

PR #1268, opened 2026-06-04
[Critical] fix: add cache size cap and inflight limit to LASA service to prevent memory leak

### LINKED ISSUE #1265: Unbounded in-memory cache in LASA service may cause memory leak under production load
## Issue

The LASA conflict detection service (`apps/api/src/services/lasa.service.ts`) uses an in-memory `Map` for caching RPC results with a 5-minute TTL, but there is no maximum size limit. Both the `cache` Map (line 35) and the `inFlight` Map (line 36) grow unbounded as users search for different medicine names.

## Cause

The `cache` at line 35 is a plain `Map<string, CacheEntry>` with no eviction strategy beyond TTL expiration. While entries do expire after 5 minutes (line 28), the Map has no cap on total entries. Under sustained production load with thousands of unique medicine name lookups per day (each unique name creates a new cache entry that persists for 5 minutes), the Map could grow to hundreds of thousands of entries. Each entry stores a `CacheEntry` object containing a `LasaMatch[]` array, which consumes increasing heap memory with no upper bound.

Similarly, the `inFlight` Map at line 36 stores in-progress promises keyed by medicine name. While entries are deleted in the `finally` block (line 89), under error conditions or aborted requests, entries could leak.

## Suggested Fix

1. Implement an LRU (Least Recently Used) eviction strategy by capping the cache at a reasonable maximum size (e.g., 1000 entries):

```typescript
const MAX_CACHE_SIZE = 1000;

function setCached(key: string, value: LasaMatch[]): void {
    if (cache.size >= MAX_CACHE_SIZE) {
        // Evict the oldest entry (first key inserted)
        const oldestKey = cache.keys().next().value;
        if (oldestKey !== undefined) {
            cache.delete(oldestKey);
        }
    }
    cache.set(key, { value, expiresAt: Date.now() + CACHE_TTL_MS });
}
```

2. Also cap the `inFlight` Map to prevent promise leaks under error conditions:

```typescript
// In the catch block or before inserting into inFlight
if (inFlight.size >= MAX_INFLIGHT) {
    inFlight.clear(); // or evict oldest
}
```

3. Consider adding a `cacheHitRate` metric or a simple counter to monitor cache effectiveness.

## Additional Context

The LASA service performs string-distance comparisons across the full medicines table via a Supabase RPC (`find_lasa_conflicts`). Under concurrent load, calling this RPC on every request without caching would exhaust the Supabase connection pool. The caching strategy is solid — the only missing piece is a size cap to prevent unbounded memory growth. The project already has the `lru-cache` or `quick-lru` patterns commonly used in Node.js, but a simple Map-based LRU is sufficient and avoids adding new dependencies.

AnushKamble:
## What issue does this fix?

Closes #1265 — The LASA conflict detection service (`apps/api/src/services/lasa.service.ts`) used an in-memory `Map` for caching RPC results with a 5-minute TTL, but had no maximum size limit. Both the `cache` Map and the `inFlight` Map could grow unbounded as users searched for different medicine names, potentially causing an OOM crash under sustained production load.

## What caused it

The `cache` Map (line 37) had no eviction strategy beyond TTL expiration. While entries expired after 5 minutes, the Map had no cap on total entries. Under sustained production load with thousands of unique medicine name lookups per day, each creating a new cache entry persisting for 5 minutes, the Map could grow to hundreds of thousands of entries. Each entry stored a `CacheEntry` object containing a `LasaMatch[]` array, consuming increasing heap memory with no upper bound.

The `inFlight` Map (line 38) also had no size cap. While entries were deleted in the `finally` block (line 97), under sustained error conditions or slow RPC responses, the Map could accumulate leaked promises.

## What the fix does

1. Added `MAX_CACHE_SIZE = 1000` constant — caps the cache at 1000 entries with LRU-like eviction (oldest entry removed when full)
2. Added `MAX_INFLIGHT = 100` constant — caps the inflight promises map, clearing it when exceeded
3. Modified `setCached()` to evict the oldest entry when the cache reaches `MAX_CACHE_SIZE` before inserting a new one
4. Added an inflight size check before inserting new promises — clears the `inFlight` map if it exceeds `MAX_INFLIGHT`

## Additional context

The LASA service performs string-distance comparisons across the full medicines table via a Supabase RPC (`find_lasa_conflicts`). Under concurrent load, calling this RPC on every request would exhaust the Supabase connection pool. The caching strategy is sound — the only missing piece was a size cap to prevent unbounded memory growth. The LRU-like eviction (deleting the first key from the Map's insertion-order iterator) is O(1) on average and sufficient for this use case without adding external dependencies.

## Tags

Please add the following labels and quality tags:
- `level:advanced`
- `quality:exceptional`
- bonus tag as appropriate

### DIFF
--- a/apps/api/src/services/lasa.service.ts
+++ b/apps/api/src/services/lasa.service.ts
@@ -26,6 +26,8 @@ export interface LasaMatch {
 //   TOCTOU window where two requests both miss the cache and both hit the DB.
 
 const CACHE_TTL_MS = 5 * 60 * 1000;
+const MAX_CACHE_SIZE = 1000;
+const MAX_INFLIGHT = 100;
 
 interface CacheEntry {
     value: LasaMatch[];
@@ -46,6 +48,12 @@ function getCached(key: string): LasaMatch[] | null {
 }
 
 function setCached(key: string, value: LasaMatch[]): void {
+    if (cache.size >= MAX_CACHE_SIZE) {
+        const oldestKey = cache.keys().next().value;
+        if (oldestKey !== undefined) {
+            cache.delete(oldestKey);
+        }
+    }
     cache.set(key, { value, expiresAt: Date.now() + CACHE_TTL_MS });
 }
 
@@ -90,6 +98,9 @@ export const detectLasaConflicts = async (medicineName: string): Promise<LasaMat
         }
     })();
 
+    if (inFlight.size >= MAX_INFLIGHT) {
+        inFlight.clear();
+    }
     inFlight.set(cacheKey, promise);
     return promise;
 };


AnushKamble:
Please review and add the following tags:
- \level:advanced\
- \quality:exceptional\
- bonus tag as appropriate

Thanks!

AnushKamble:
Please @dipexplorer, can you add the following tag, as it took a lot of effort constructing this PR and all:

\level:intermediate\
\quality:exceptional\
bonus tag as appropriate
Thanks!

PR #1325, opened 2026-06-05
[Critical] fix(api): add logging to geocodePincode silent catch block (#1315)

### LINKED ISSUE #1315: [Critical] geocodePincode() in api.ts silently swallows errors without logging
## Issue
`apps/web/lib/api.ts` lines 140-142 define a `geocodePincode()` function that catches all errors and returns `null` with zero logging:

```typescript
} catch {
    return null;
}
```

When geocoding fails — whether due to network timeout, DNS failure, malformed pincode, or Nominatim rate limiting — callers receive `null` silently. This makes it extremely difficult to diagnose why a pincode lookup failed, especially in production environments where console output is not visible.

## Cause
The catch block was designed for graceful degradation (if geocoding fails, the report can still be submitted without coordinates). However, swallowing the error entirely removes all observability. When a user reports that their address isn't being geocoded correctly, there is no way to distinguish between:

- Network connectivity issues
- Nominatim API rate limiting (1 req/sec)
- Invalid pincode format
- Server-side errors from Nominatim
- Abort signal triggered timeout

The calling code in `submitReport()` (api.ts line 80) receives `null` for coordinates and proceeds, which is acceptable fallback behavior — but the lack of logging means operators cannot distinguish between "user has poor connectivity" from "Nominatim is rate limiting us" from "user entered an invalid pincode".

## Suggested Fix
Add descriptive logging using the existing `structuredLog` utility before returning null:

```typescript
} catch (error) {
    // Log the error for observability, then gracefully degrade
    if (typeof window !== 'undefined') {
        console.warn(`[api] Geocoding pincode ${pincode} failed: ${error instanceof Error ? error.message : String(error)}`);
    }
    return null;
}
```

Or using the structured logger if available in the context:

```typescript
} catch (error) {
    logger.warn('[api] Geocoding pincode failed', {
        pincode,
        error: error instanceof Error ? error.message : String(error),
    });
    return null;
}
```

The key is that the error message should include the pincode (masked if necessary for PII) so that support requests can correlate the failure with specific user reports.

## Additional Context
- `geocodePincode` is called from `submitReport()` to optionally enrich reports with GPS coordinates
- The Nominatim usage policy requires max 1 request/second and recommends identifying the application
- The function already has a 4-second timeout via `AbortSignal.timeout(4000)` at line 123
- A more robust solution would be to cache successful geocodes in sessionStorage to avoid repeated Nominatim calls for the same pincode, reducing API load and improving UX
- Future enhancement: consider using a Supabase database edge function for server-side geocoding to avoid client-side rate limiting

AnushKamble:
## Issue
`apps/web/lib/api.ts` lines 140-142 defined a `geocodePincode()` function that caught all errors and returned `null` with zero logging. When geocoding failed — whether due to network timeout, DNS failure, malformed pincode, or Nominatim rate limiting — callers received `null` silently, making it extremely difficult to diagnose failures in production.

## Cause
The catch block was designed for graceful degradation (if geocoding fails, the report can still be submitted without coordinates). However, swallowing the error entirely removed all observability. Operators could not distinguish between "user has poor connectivity", "Nominatim is rate limiting us", or "user entered an invalid pincode".

Before:
```
} catch {
    return null;
}
```

## Fix
Added a parameter to the catch and descriptive logging before returning null:

After:
```
} catch (error) {
    if (typeof window !== 'undefined') {
        console.warn(`[api] Geocoding pincode ${pincode} failed: ${error instanceof Error ? error.message : String(error)}`);
    }
    return null;
}
```

## Additional Context
- `geocodePincode` is called from `submitReport()` to optionally enrich reports with GPS coordinates
- The Nominatim usage policy requires max 1 request/second and recommends identifying the application
- The function already had a 4-second timeout via `AbortSignal.timeout(4000)` at line 123
- This is a low-risk observability fix — the `return null` fallback behavior is unchanged, only logging is added

## Closes
Closes #1315

### DIFF
--- a/apps/web/lib/api.ts
+++ b/apps/web/lib/api.ts
@@ -137,7 +137,10 @@ export async function geocodePincode(
         const lng = parseFloat(arr[0].lon);
         if (!Number.isFinite(lat) || !Number.isFinite(lng)) return null;
         return { latitude: lat, longitude: lng };
-    } catch {
+    } catch (error) {
+        if (typeof window !== "undefined") {
+            console.warn(`[api] Geocoding pincode ${pincode} failed: ${error instanceof Error ? error.message : String(error)}`);
+        }
         return null;
     }
 }


AnushKamble:
Hey @AnushKamble! The PR looks good. Could you please review and add the following tags to this PR if it meets the criteria:
- intermediate
- exceptional
- onus

Thank you!

AnushKamble:
@dipexplorer,
I noticed level, quality, type tags were not added to this PR which are required for GSsoC Scoring.
Kindly add the appropriate tags.
Thanks!

PR #1507, opened 2026-06-07
fix: include full last day in SafetyStatsBanner monthly query

### LINKED ISSUE #1501: [BUG] SafetyStatsBanner excludes most alerts from the last day of the month
### Describe the bug
`SafetyStatsBanner` labels its counts as the current month, but the Supabase date range ends at midnight at the start of the month's final day. Alerts created later on the last calendar day are excluded from the monthly totals.

### Source proof
In `apps/web/components/SafetyStatsBanner.tsx`:

```ts
const startOfMonth = new Date(now.getFullYear(), now.getMonth(), 1).toISOString();
const endOfMonth = new Date(now.getFullYear(), now.getMonth() + 1, 0).toISOString();
...
.gte("created_at", startOfMonth)
.lte("created_at", endOfMonth);
```

`new Date(year, month + 1, 0)` creates the last day at `00:00:00`, not the end of that day. For example, a June 30 alert at 10:00 is greater than the `endOfMonth` value and is not counted.

### Expected behavior
The end boundary should include the full last day, or the query should use an exclusive first-day-of-next-month upper bound.

This is a date-boundary bug visible directly in the query construction.

vipul674:
Closes #1501

Uses an exclusive `startOfNextMonth` upper bound instead of `endOfMonth` (which resolved to midnight of the last day), so alerts created later on the final calendar day are correctly included in the monthly totals.

### DIFF
--- a/apps/web/components/SafetyStatsBanner.tsx
+++ b/apps/web/components/SafetyStatsBanner.tsx
@@ -201,13 +201,17 @@ export default function SafetyStatsBanner() {
         async function fetchAlerts() {
             const now = new Date();
             const startOfMonth = new Date(now.getFullYear(), now.getMonth(), 1).toISOString();
-            const endOfMonth = new Date(now.getFullYear(), now.getMonth() + 1, 0).toISOString();
+            const startOfNextMonth = new Date(
+                now.getFullYear(),
+                now.getMonth() + 1,
+                1
+            ).toISOString();
 
             const { data, error } = await supabase
                 .from("drug_alerts")
                 .select("alert_type")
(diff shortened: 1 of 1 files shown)


vipul674:
@RatLoopz Ready for review.

## What changed

SafetyStatsBanner was using `endOfMonth = new Date(year, month + 1, 0)` which resolves to midnight of the last calendar day, not the end of that day. Alerts created later on the last day were silently excluded from the monthly count.

## Fix

Replaced the upper bound with an exclusive `startOfNextMonth` and switched from `.lte` to `.lt`, so every timestamp within the calendar month is captured.

Closes #1501

PR #1508, opened 2026-06-07
fix: handle overnight operating hours in PharmacyStatusBadge

### LINKED ISSUE #1500: [BUG] Pharmacy open badge fails for overnight operating hours
### Describe the bug
`PharmacyStatusBadge` treats every operating-hours range as same-day. Pharmacies that operate overnight, such as `21:00 - 06:00`, are shown as closed during their after-midnight open period.

### Source proof
In `apps/web/components/PharmacyCard.tsx`, `isPharmacyOpen()` parses open/close minutes and returns:

```ts
return currentMinutes >= openMinutes && currentMinutes < closeMinutes;
```

For an overnight schedule like `21:00 - 06:00`, `openMinutes` is greater than `closeMinutes`, so no time after midnight can satisfy both comparisons. At 01:00, the function returns false even though the pharmacy should still be open.

### Expected behavior
When the closing time is earlier than the opening time, the check should treat the range as crossing midnight, e.g. current time is open when it is after opening OR before closing.

The bug is a deterministic time-range logic error and can be verified with a unit test for `isPharmacyOpen("21:00 - 06:00")`.

vipul674:
Closes #1500

When the closing time is earlier than the opening time (e.g. `21:00 - 06:00`), the range now correctly treats it as crossing midnight. The pharmacy shows as open if the current time is after opening OR before closing.

### DIFF
--- a/apps/web/components/PharmacyCard.tsx
+++ b/apps/web/components/PharmacyCard.tsx
@@ -19,7 +19,10 @@ export function isPharmacyOpen(operatingHours: string): boolean {
     const openMinutes = openH * 60 + openM;
     const closeMinutes = closeH * 60 + closeM;
 
-    return currentMinutes >= openMinutes && currentMinutes < closeMinutes;
+    if (openMinutes < closeMinutes) {
+        return currentMinutes >= openMinutes && currentMinutes < closeMinutes;
+    }
+    return currentMinutes >= openMinutes || currentMinutes < closeMinutes;
 }
 
 interface PharmacyStatusBadgeProps {


vipul674:
@RatLoopz Ready for review.

## What changed

isPharmacyOpen() assumed all operating-hour ranges were same-day. Overnight ranges like `21:00 - 06:00` always returned false because openMinutes > closeMinutes made the comparison impossible.

## Fix

When closeMinutes <= openMinutes the function now treats the range as crossing midnight: open if current time is after opening OR before closing.

Closes #1500

PR #1509, opened 2026-06-07
fix: validate radius_km in nearby map API route

### LINKED ISSUE #1502: [BUG] Nearby map API accepts invalid radius values and passes them to the RPC
### Describe the bug
`GET /api/map/nearby` validates that latitude and longitude are numbers, but it does not validate `radius_km`. Invalid, zero, or negative radius values can be passed directly into the `get_nearest_pharmacies` RPC.

### Source proof
In `apps/api/src/routes/map.ts`:

```ts
const radius_km = parseFloat((req.query.radius_km as string) || "10");

if (isNaN(lat) || isNaN(lng)) {
  return res.status(400).json({ error: "lat and lng are required query params" });
}
...
search_radius_km: radius_km,
```

There is no `Number.isFinite(radius_km)` check and no positive-range check. Requests like `/api/map/nearby?lat=18.52&lng=73.85&radius_km=abc` or `radius_km=-5` reach the database RPC instead of returning a controlled 400.

### Expected behavior
The route should validate `radius_km` as a finite positive number and clamp it to an acceptable maximum before calling Supabase.

This is an API validation bug visible in the route logic and reproducible with a malformed query string.

vipul674:
Closes #1502

Validates `radius_km` as a finite positive number and clamps it to a maximum of 100 km before passing it to the `get_nearest_pharmacies` RPC. Invalid values (`NaN`, `Infinity`, zero, negative) now return a controlled 400 error.

### DIFF
--- a/apps/api/src/routes/map.ts
+++ b/apps/api/src/routes/map.ts
@@ -47,11 +47,16 @@ router.get("/nearby", async (req: Request, res: Response) => {
         return res.status(400).json({ error: "lat and lng are required query params" });
     }
 
+    if (!Number.isFinite(radius_km) || radius_km <= 0) {
+        return res.status(400).json({ error: "radius_km must be a positive number" });
+    }
+    const clampedRadius = Math.min(radius_km, 100);
+
     try {
         const pharmaciesRes = await supabase.rpc("get_nearest_pharmacies", {
             query_lat: lat,
             query_lng: lng,
-            search_radius_km: radius_km,
+            search_radius_km: clampedRadius,
         });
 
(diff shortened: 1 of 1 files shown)


vipul674:
@RatLoopz Ready for review.

## What changed

GET /api/map/nearby parsed radius_km from the query string but never validated it. Non-finite, zero, or negative values were forwarded directly to the Supabase RPC.

## Fix

Added a `.isFinite` + positive-range check that returns 400 for bad values, and clamps valid radii to a 100 km ceiling before calling the RPC.

Closes #1502

PR #1561, opened 2026-06-09
fix: ignore stale search responses in MedicineSearchSelect

### LINKED ISSUE #1503: [BUG] Medicine comparison search can show stale results from an older query
### Describe the bug
The medicine comparison combobox does not cancel or guard against older in-flight searches. If a slower search for an earlier query resolves after a newer query, it can overwrite the results list with stale medicines.

### Source proof
In `apps/web/src/components/MedicineSearchSelect.tsx`:

- The debounced `useEffect` calls `const res = await onSearch(q)`.
- It then unconditionally calls `setResults(res)` and `setActiveIndex(...)`.
- There is no request id, abort controller, or check that the current `query.trim()` still equals the `q` captured by that request.

So a slow response for `pa` can replace the results for a later query such as `para`, causing the user to select from the wrong result set.

### Expected behavior
Only the latest active search response should update the combobox. Older responses should be ignored or aborted.

This is an asynchronous state race visible from the component logic and can be verified with mocked delayed `onSearch` responses.

vipul674:
## Summary

- Fixes a race condition where a slow search response for an earlier query could overwrite results for a newer query
- Adds a ref to track the latest query and skip updating state when the response no longer matches

## Files Changed

- `apps/web/src/components/MedicineSearchSelect.tsx`: Added `latestQueryRef` to guard against stale search responses

## Testing Performed

- `npm ci`
- `npm run build`

## AI Assistance Disclosure

This PR was prepared with AI assistance and manually reviewed before submission.

Closes #1503


vipul674:
@dipexplorer PR is ready for review.

## Report

### Issue
The debounced search effect in `MedicineSearchSelect` did not guard against older in-flight requests. A slow response for an earlier query (e.g., `pa`) could overwrite results for a newer query (e.g., `para`), showing the user stale medicines.

### Changes
- `apps/web/src/components/MedicineSearchSelect.tsx`: Added `latestQueryRef` to track the current query

### Fix Details
- Added `useRef` to track the latest query string
- Set `latestQueryRef.current = q` before starting the debounced search
- After `onSearch` returns, check if `latestQueryRef.current !== q` — if a newer query was issued, skip updating state
- The stale response is silently discarded, preventing stale results from appearing

### Files Changed
- `apps/web/src/components/MedicineSearchSelect.tsx` (+3, -0)

### Verification
- Race condition fixed: slow responses no longer overwrite newer results
- Existing search behavior unchanged for normal (non-racing) queries
- No AbortController needed — simple ref check is sufficient and avoids complexity

PR #1609, opened 2026-06-09
[Critical] Predictable CSRF secret fallback in production — no configuration guard

### LINKED ISSUE #1606: [Critical] Predictable CSRF secret fallback in production — no configuration guard
## Description

The CSRF token signing secret in `apps/api/src/app.ts` falls back to a hardcoded predictable string when `CSRF_SECRET` environment variable is not set. Unlike `SUPABASE_URL` and `SUPABASE_ANON_KEY` which have startup guards (lines 20-28), there is no check for `CSRF_SECRET`, so the app silently runs with a compromised secret in every environment.

## Cause

At line 70, the fallback `||` operator substitutes a well-known string:

```typescript
getSecret: () => process.env.CSRF_SECRET || "fallback-secret-change-in-production",
```

An attacker who knows this string can forge valid CSRF tokens for any state-changing endpoint, completely bypassing CSRF protection.

## Fix

Add a startup guard that crashes the process if `CSRF_SECRET` is missing in non-development environments, matching the existing pattern for `SUPABASE_URL`:

```typescript
if (process.env.NODE_ENV !== "development" && !process.env.CSRF_SECRET) {
    console.error("Missing CSRF_SECRET environment variable.");
    process.exit(1);
}
```

## Additional Context

Vulnerable endpoints include: `POST /api/reports`, `POST /api/verify`, `POST /api/v1/scan/*`, `POST /api/v1/lasa/check`, `POST /api/notifications/*`, `PATCH /api/reports/:id/status`, and all `POST /api/v1/admin/*` routes.

## Follow-up

Security fix. The application should never silently run with a predictable CSRF secret in staging or production.

AnushKamble:
## Description
#1606 — The CSRF token signing secret in apps/api/src/app.ts:70 falls back to the hardcoded predictable string "fallback-secret-change-in-production" when CSRF_SECRET is not set. An attacker who knows this string can forge valid CSRF tokens for any state-changing endpoint.

## Cause
Line 70: getSecret: () => process.env.CSRF_SECRET || "fallback-secret-change-in-production"

Unlike SUPABASE_URL and SUPABASE_ANON_KEY (which have explicit startup guards at lines 20-28 that call process.exit(1)), there is no check for CSRF_SECRET. The app starts normally with a compromised CSRF secret.

## Fix
Added a startup guard matching the existing pattern:

```typescript
if (process.env.NODE_ENV !== "development" && !process.env.CSRF_SECRET) {
    logger.error("Missing CSRF_SECRET environment variable.");
    process.exit(1);
}
```

## Vulnerable Endpoints
POST /api/reports, POST /api/verify, POST /api/v1/scan/*, POST /api/v1/lasa/check, POST /api/notifications/*, PATCH /api/reports/:id/status, all POST /api/v1/admin/*

## Additional Context
Security fix. The application should never silently run with a predictable CSRF secret in staging or production.

Closes #1606


AnushKamble:
@RatLoopz Branch is rebased against latest upstream/main and ready for merging.

PR #2377, opened 2026-06-22
fix(api): build single .or() expression for scan/match fallback

### LINKED ISSUE #2374: [BUG] /scan/match fallback search only matches last word due to .or() overwrite
## Bug Description

In the `POST /api/v1/scan/match` route, the fallback database search calls `.or()` in a loop. Each call to `.or()` in the Supabase JS client overwrites the previous filter — only the last word's filter actually applies. Multi-word search queries silently fail to match earlier words.

**File:** `apps/api/src/routes/scan.ts` — the fallback path inside `/match`

## Expected Behavior

When a user searches "paracetamol 500mg tablet", the fallback should match medicines containing **any** of the words: paracetamol, 500mg, tablet. Currently it only matches the last word ("tablet").

## Steps to Reproduce

1. Start the API server locally
2. Ensure no exact match exists in the medicines table for a multi-word query
3. Send `POST /api/v1/scan/match` with body `{ "query": "paracetamol 500mg" }`
4. Observe the fallback only searches for "500mg" — "paracetamol" is ignored

## Code Evidence

```typescript
// apps/api/src/routes/scan.ts — fallback path
let fallbackQuery = supabase.from("medicines").select("brand_name, generic_name");

for (const word of words) {
  fallbackQuery = fallbackQuery.or(
    `brand_name.ilike.%${escapePostgrest(word)}%,generic_name.ilike.%${escapePostgrest(word)}%`
  );
}
```

In the Supabase JS client, `.or()` writes to a single `or` query parameter. Each iteration overwrites the previous value. After the loop, only the last word's filter remains active.

## Suggested Fix

Build the full `.or()` expression as a single string before passing it to Supabase:

```typescript
const orConditions = words.map(word =>
  `(brand_name.ilike.%${escapePostgrest(word)}%,generic_name.ilike.%${escapePostgrest(word)}%)`
).join(",");

const { data: fallback } = await supabase
  .from("medicines")
  .select("brand_name, generic_name")
  .or(orConditions)
  .limit(3);
```

vipul674:
## Summary

Fixes a bug in `POST /api/v1/scan/match` where the fallback database search called `.or()` in a loop. Each call overwrites the previous filter in the Supabase JS client, so multi-word queries silently only matched the last word.

## Changes

- **`apps/api/src/routes/scan.ts`**: Build the full OR expression as a single string before passing it to Supabase, instead of calling `.or()` in a loop.

## Impact

**Before:** Searching "paracetamol 500mg" would only search for medicines matching "500mg".

**After:** Searching "paracetamol 500mg" searches for medicines matching either "paracetamol" OR "500mg".

## Closes

Closes #2374

### DIFF
--- a/apps/api/src/routes/scan.ts
+++ b/apps/api/src/routes/scan.ts
@@ -673,15 +673,18 @@ router.post("/match", scanQueryLimiter, async (req: Request, res: Response) => {
                 .split(/\s+/)
                 .filter((w: string) => w.length > 2);
             if (words.length > 1) {
-                let fallbackQuery = supabase.from("medicines").select("brand_name, generic_name");
+                const orConditions = words
+                    .map(
+                        (w: string) =>
+                            `brand_name.ilike.%${escapePostgrest(w)}%,generic_name.ilike.%${escapePostgrest(w)}%`
+                    )
+                    .join(",");
 
-                for (const word of words) {
-                    fallbackQuery = fallbackQuery.or(
-                        `brand_name.ilike.%${escapePostgrest(word)}%,generic_name.ilike.%${escapePostgrest(word)}%`
-                    );
-                }
-
-                const { data: fallback } = await (fallbackQuery as any).limit(3);
+                const { data: fallback } = await supabase
+                    .from("medicines")
+                    .select("brand_name, generic_name")
+                    .or(orConditions)
+                    .limit(3);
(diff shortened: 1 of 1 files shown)


vipul674:
@dipexplorer This PR is ready for review. It fixes the scan/match fallback where .or() was called in a loop, only matching the last word. TypeScript compiles clean. cc @vipul674

PR #2397, opened 2026-06-22
fix(api): prevent escapePostgrest from escaping ILIKE wildcards in medicine RAG fallback

### LINKED ISSUE #2395: [BUG] Medicine RAG Fallback Search Broken — escapePostgrest() Double-Escapes ILIKE Wildcard %, Returns 0 Results

### Bug Description
In `apps/api/src/services/medicineRag.service.ts` (the fallback medicine search), the `escapePostgrest()` function is applied to the **full ILIKE pattern including outer `%` wildcards**, which escapes those wildcards into literal `%` characters. This breaks the fallback search (Tier 3) when the primary RPC search is unavailable.

**Root cause** — lines 286–293:

```ts
const pattern = `%${escapeIlike(query)}%`;
const { data: tableData, error: tableError } = await anonSupabase
    .from("medicines")
    .select("...")
    .or(
        `generic_name.ilike."${escapePostgrest(pattern)}",brand_name.ilike."${escapePostgrest(pattern)}",composition.ilike."${escapePostgrest(pattern)}"`
    )
    .limit(limit);
```

The `pattern` variable already contains the outer `%...%` wildcards needed by ILIKE. But `escapePostgrest()` replaces `%` → `\%` globally, which escapes those wildcard `%` characters too.

**What happens at runtime:**

- User searches for `"paracetamol"`
- `pattern` = `"%paracetamol%"`
- `escapePostgrest("%paracetamol%")` → `"\%paracetamol\%"`
- In PostgreSQL ILIKE syntax, `\%` is a **literal** `%` character, **not** a wildcard
- The query searches for rows whose `generic_name` literally contains `%paracetamol%` — which is virtually none
- **Result:** the fallback search returns 0 results even when matching medicines exist

**Contrast with correct usage** — other routes apply `escapePostgrest()` only to the value, not the wildcards:

```ts
// alternatives.ts (correct):
`brand_name.ilike."%${escapePostgrest(brandName)}%"`     // outer % stay as wildcards

// scan.ts (correct):
`brand_name.ilike.%${escapePostgrest(name)}%`              // outer % stay as wildcards

// medicineRag.service.ts (BROKEN):
pattern = "%" + value + "%"    // outer % included BEFORE escapePostgrest
escapePostgrest(pattern)        // escapes the outer % too!
```

**Impact:**

The Tier-3 fallback occurs when the primary RPC (`search_medicines_text`) and Tier-2 (`ts_query`) are both unavailable (e.g., Supabase connection pool exhausted, network blips, or missing pg_trgm extension). In this degraded state, any medicine brand/generic search via the triage chatbot or medicine lookup interface returns zero results, making the entire search feature silently broken.

<small>Tracing: `searchMedicines` → tries RPC → fails → tries `ts_query` → fails → enters Tier-3 ILIKE fallback → broken pattern → returns `[]`.</small>

### Expected Behavior
The ILIKE wildcards (`%`) should not be escaped. The fix is to apply `escapePostgrest()` **before** wrapping with outer `%`, or to build the pattern differently:

```ts
// Fix: escape value first, then wrap with wildcards
const safe = escapePostgrest(escapeIlike(query));
const orFilter = `generic_name.ilike."%${safe}%",brand_name.ilike."%${safe}%",composition.ilike."%${safe}%"`;
```

This matches the pattern used by `alternatives.ts` and `scan.ts`.

### Screenshots
N/A

### GSSoC 2026

### Code of Conduct
- [x] I have read and agree to SahiDawa's Code of Conduct.

vipul674:
## Summary

Fixes #2395 — medicineRag.service.ts Tier-3 fallback search returns 0 results because `escapePostgrest()` double-escapes the outer ILIKE `%` wildcards.

## Root Cause

The `pattern` variable was built with outer `%` wildcards first, then `escapePostgrest()` was applied — which escapes `%` globally, turning the wildcards into literal `\%` characters.

```ts
// Before (broken)
const pattern = `%${escapeIlike(query)}%`;
.or(`...ilike."${escapePostgrest(pattern)}"...`)  // outer % escaped too!
```

## Fix

Escape the value `before` wrapping with outer `%` wildcards:

```ts
// After (fixed)
const safe = escapePostgrest(escapeIlike(query));
const pattern = `%${safe}%`;
.or(`...ilike."${pattern}"...`)
```

This matches the pattern used by `alternatives.ts` and `scan.ts`.

### DIFF
--- a/apps/api/src/services/medicineRag.service.ts
+++ b/apps/api/src/services/medicineRag.service.ts
@@ -283,14 +283,15 @@ export async function retrieveRelevantMedicines(
     });
 
     // Tier 3 — in-memory ILIKE filter over the table.
-    const pattern = `%${escapeIlike(query)}%`;
+    const safe = escapePostgrest(escapeIlike(query));
+    const pattern = `%${safe}%`;
     const { data: tableData, error: tableError } = await anonSupabase
         .from("medicines")
         .select(
             "id, brand_name, generic_name, manufacturer, composition, strength, dosage_form, schedule, mrp, jan_aushadhi_price"
         )
         .or(
-            `generic_name.ilike."${escapePostgrest(pattern)}",brand_name.ilike."${escapePostgrest(pattern)}",composition.ilike."${escapePostgrest(pattern)}"`
+            `generic_name.ilike."${pattern}",brand_name.ilike."${pattern}",composition.ilike."${pattern}"`
         )
         .limit(limit);
 


vipul674:
@dipexplorer ready for review

PR #2411, opened 2026-06-23
[BUG] ML Service Crashes on Startup — cdsco_alert_agent.py Calls sys.exit(1) at Module Import Time

### LINKED ISSUE #2408: [BUG] ML Service Crashes on Startup — cdsco_alert_agent.py Calls sys.exit(1) at Module Import Time

## Bug Description

Importing `apps/ml/agent/cdsco_alert_agent.py` at module level calls `sys.exit(1)` if `API_BASE_URL` or `API_SECRET_KEY` environment variables are not set. This happens unconditionally outside `if __name__ == "__main__":`, so **any code path that imports this module — even just to access a helper function — will kill the entire Python process** with exit code 1 and no traceback.

### Evidence

`apps/ml/agent/cdsco_alert_agent.py`, lines 19–22 and 25–28:

```python
API_BASE_URL = os.getenv("API_BASE_URL", "").strip().rstrip("/")
if not API_BASE_URL:
    logging.error("CRITICAL ERROR: API_BASE_URL is not set in environment.")
    sys.exit(1)

API_SECRET_KEY = os.getenv("API_SECRET_KEY")
if not API_SECRET_KEY:
    logging.error("CRITICAL ERROR: API_SECRET_KEY is not set in environment.")
    sys.exit(1)
```

This code is **outside** `if __name__ == "__main__":` (which starts at line 151), so it runs every time the module is imported.

### How this reaches production

Although the main FastAPI app does not import this module directly, the script is designed to be run as a standalone cron/agent process. Any test runner, import-based utility, or downstream script that does `import cdsco_alert_agent` will crash silently.

## Expected Behavior

Importing the module should not terminate the process. Missing configuration should raise a recoverable exception (e.g., `RuntimeError`) or leave variables as `None` and let callers handle missing config.

## Actual Behavior

```
$ python3 -c "import cdsco_alert_agent"
CRITICAL ERROR: API_BASE_URL is not set in environment.
$ echo $?
1
```
The process terminates immediately with no way to catch or handle the error.

## Affected files

- `apps/ml/agent/cdsco_alert_agent.py` lines 18–28 (module-level sys.exit)

## Suggested Fix

Move the environment variable validation + `sys.exit(1)` inside the `if __name__ == "__main__":` block. For import-time safety, raise a `RuntimeError` instead, or let the variables default to `None` so consumers can handle missing config gracefully:

```python
API_BASE_URL = os.getenv("API_BASE_URL", "").strip().rstrip("/")
API_SECRET_KEY = os.getenv("API_SECRET_KEY")
# Remove sys.exit(1) — allow consumers to handle missing config
```

Then at script entry point (line 151):
```python
if __name__ == "__main__":
    if not API_BASE_URL:
        logging.error("API_BASE_URL is not set. Exiting.")
        sys.exit(1)
    scrape_cdsco_alerts()
```

## Impact

- **Process termination**: Any import of this module kills the process unconditionally
- **Silent failure**: No recoverable exception, just exit code 1
- **Developers blocked**: Importing the agent module for testing or reuse is impossible without setting env vars first
- **Hard to debug**: The error message says "CRITICAL ERROR" but gives no traceback, making root cause analysis harder than necessary

## GSSoC 2026

## Code of Conduct

vipul674:
## Description

Move  calls out of module-level code into the  guard so importing the module doesn't kill the Python process.

### Changes
- Remove  from module-level env var validation in 
- Derive  and  lazily (only when base URL is set)
- Add env var check inside  block before calling 

Fixes #2408

### DIFF
--- a/apps/ml/agent/cdsco_alert_agent.py
+++ b/apps/ml/agent/cdsco_alert_agent.py
@@ -16,18 +16,10 @@ logging.basicConfig(level=logging.INFO)
 CDSCO_ALERTS_URL = "https://cdsco.gov.in/opencms/opencms/en/Notifications/Alerts/"
 
 API_BASE_URL = os.getenv("API_BASE_URL", "").strip().rstrip("/")
-if not API_BASE_URL:
-    logging.error("CRITICAL ERROR: API_BASE_URL is not set in environment.")
-    sys.exit(1)
-
-INGEST_API_URL = API_BASE_URL + "/api/v1/alerts/ingest"
-
 API_SECRET_KEY = os.getenv("API_SECRET_KEY")
-if not API_SECRET_KEY:
-    logging.error("CRITICAL ERROR: API_SECRET_KEY is not set in environment.")
-    sys.exit(1)
 
-ALERTS_API_URL = API_BASE_URL + "/api/v1/alerts"
+INGEST_API_URL = API_BASE_URL + "/api/v1/alerts/ingest" if API_BASE_URL else ""
+ALERTS_API_URL = API_BASE_URL + "/api/v1/alerts" if API_BASE_URL else ""
 
 def scrape_cdsco_alerts():
     logging.info(f"Checking {CDSCO_ALERTS_URL} for new alerts...")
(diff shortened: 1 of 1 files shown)


vipul674:
@dipexplorer @Er-ine @oriazephyr @Siddesh-G-H PR ready for review. Moves sys.exit(1) out of module-level code into __main__ guard so importing cdsco_alert_agent doesn't kill the process.

PR #2412, opened 2026-06-23
[BUG] ML Service Tracing Config from .env Silently Ignored — setup_tracing() Called Before load_dotenv()

### LINKED ISSUE #2409: [BUG] ML Service Tracing Config from .env Silently Ignored — setup_tracing() Called Before load_dotenv()

## Bug Description

In `apps/ml/main.py`, `setup_tracing()` is called at **line 11**, but `load_dotenv()` is called at **line 17**. Inside `setup_tracing()` (defined in `apps/ml/tracing.py`, line 9), the function reads `os.getenv("OTEL_EXPORTER_OTLP_ENDPOINT")` to configure the OpenTelemetry OTLP exporter. If this variable is set in a `.env` file (which is standard practice for local development and deployment), it has **not yet been loaded** when `setup_tracing()` runs.

The result: the OTLP exporter is never configured, and OpenTelemetry tracing silently does nothing with no warning or error logged.

### Evidence

`apps/ml/main.py` (lines 11 and 17):

```python
setup_tracing()                          # line 11 — reads env before .env is loaded
RequestsInstrumentor().instrument()

from services.telemetry import configure_telemetry_logging
from services.router_loader import include_router_if_available

load_dotenv()                            # line 17 — .env loaded too late
configure_telemetry_logging()
```

`apps/ml/tracing.py` (lines 8-10):

```python
def setup_tracing():
    endpoint = os.getenv("OTEL_EXPORTER_OTLP_ENDPOINT")   # reads None if set in .env
```

### Environment context

The project's `.env.example` does not include `OTEL_EXPORTER_OTLP_ENDPOINT`, but in production deployments it is standard to configure tracing via .env or deployment secrets. Engineers who set tracing configuration in `.env` will find it silently ignored.

## Expected Behavior

If `OTEL_EXPORTER_OTLP_ENDPOINT` is set in `.env`, `setup_tracing()` should read it and configure the exporter correctly.

## Actual Behavior

`setup_tracing()` always reads `None` for the endpoint when the value is in `.env`, so tracing is never set up. No error or warning is logged.

## Affected files

- `apps/ml/main.py` — line 11 vs line 17 (ordering)
- `apps/ml/tracing.py` — line 9 (`os.getenv` reads system env only)

## Suggested Fix

Move `setup_tracing()` to after `load_dotenv()`:

```python
load_dotenv()
configure_telemetry_logging()
setup_tracing()             # Moved here — .env is now loaded
RequestsInstrumentor().instrument()
```

## Impact

- **Tracing silently broken**: Engineers setting up observability via .env will see no traces and no error message
- **Hard to debug**: The cause is a simple ordering issue, but there is no logged warning to hint at the problem
- **Lost visibility**: In production, missing traces means engineers lose critical debugging capability during incidents

## GSSoC 2026

## Code of Conduct

vipul674:
## Description

Reorder startup sequence so `load_dotenv()` runs before `setup_tracing()`. The tracing function reads `OTEL_EXPORTER_OTLP_ENDPOINT` from environment, but if the value is set in `.env` it was silently ignored because `load_dotenv()` hadn't run yet.

### Changes
- Move `load_dotenv()` call to before `setup_tracing()` in `apps/ml/main.py`

Fixes #2409

### DIFF
--- a/apps/ml/main.py
+++ b/apps/ml/main.py
@@ -8,13 +8,14 @@ from tracing import setup_tracing
 from opentelemetry.instrumentation.fastapi import FastAPIInstrumentor
 from opentelemetry.instrumentation.requests import RequestsInstrumentor
 
+load_dotenv()
+
 setup_tracing()
 RequestsInstrumentor().instrument()
 
 from services.telemetry import configure_telemetry_logging
 from services.router_loader import include_router_if_available
 
-load_dotenv()
 configure_telemetry_logging()
 logger = logging.getLogger(__name__)
 


vipul674:
@dipexplorer @Er-ine @oriazephyr @Siddesh-G-H PR ready for review. Reorders startup so load_dotenv() runs before setup_tracing(), fixing silent .env config ignore.

PR #2456, opened 2026-06-24
[GSSoC'26] fix(api): allow CORS requests from non-browser clients

### LINKED ISSUE #2452: [BUG] CORS Middleware Blocks All Non-Browser API Clients — Server-to-Server, cURL, Mobile Apps Cannot Reach Any Endpoint

### Bug Description

**Location:** `apps/api/src/config/cors.ts` (lines 39-45)

The CORS configuration rejects all requests where the `Origin` header is undefined or falsy. This blocks ALL non-browser API consumers — including server-to-server calls, mobile apps, cURL, Postman, and any programmatic client that does not send an `Origin` header.

**Root Cause:** The `origin` callback function in `createCorsOptions()` uses:

```typescript
origin(origin, callback) {
    if (origin && allowedOrigins.includes(origin)) {
        callback(null, true);
        return;
    }
    callback(null, false);  // ← fires when origin is undefined
}
```

When `origin` is `undefined` (which is the standard for non-browser HTTP clients), the condition `origin && allowedOrigins.includes(origin)` short-circuits to `false`, and `callback(null, false)` blocks the request entirely.

The `cors` package interprets `false` as a denial and omits the `Access-Control-Allow-Origin` header, causing browsers (and servers) to reject the response.

This is especially impactful because the API's own `/health` endpoint on the root route (app.ts:176) advertises the API as having a `/health` endpoint, but that endpoint is equally unreachable from non-browser clients.

**Evidence:**

- CORS middleware is applied globally at `app.ts:151` via `app.use(cors(createCorsOptions()))`
- Every API route is affected because CORS is registered before all routes
- The `trust proxy` setting at `app.ts:78` indicates the API already expects non-browser proxied traffic

### Expected Behavior

Non-browser clients (server-to-server, curl, mobile) should be allowed through CORS when they send no `Origin` header. The fix is:

```typescript
origin(origin, callback) {
    if (!origin || allowedOrigins.includes(origin)) {
        callback(null, true);
        return;
    }
    callback(null, false);
}
```

This allows requests without an `Origin` header (non-browser clients) while still rejecting unknown origins from browsers.

For additional security, consider adding the `Vary: Origin` response header to ensure CDNs and proxies cache responses correctly per origin.

### Screenshots

_No response_

### GSSoC 2026

### Code of Conduct

vipul674:
## Description

- Fixed CORS middleware to allow requests without an `Origin` header (non-browser clients)
- Changed `if (origin && allowedOrigins.includes(origin))` to `if (!origin || allowedOrigins.includes(origin))`
- Non-browser clients (curl, Postman, server-to-server, mobile apps) now pass through CORS without being blocked

## Files Changed

- `apps/api/src/config/cors.ts`: Changed origin check to allow undefined origins

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation

## Difficulty & Label Request

Assessed difficulty: `level1`

Please apply the matching difficulty label for GSSoC '26 scoring.
If applicable, please also apply `gssoc:approved` after review.

## Testing & Verification

Commands run:

- Code review: verified logic change allows undefined origin while still rejecting unknown origins

## Known Limitations

- None

## GSSoC 2026 Compliance

This contribution was prepared with AI assistance and manually reviewed before submission.

Closes #2452

### DIFF
--- a/apps/api/src/config/cors.ts
+++ b/apps/api/src/config/cors.ts
@@ -37,7 +37,7 @@ export function createCorsOptions(env: NodeJS.ProcessEnv = process.env): CorsOpt
 
     return {
         origin(origin, callback) {
-            if (origin && allowedOrigins.includes(origin)) {
+            if (!origin || allowedOrigins.includes(origin)) {
                 callback(null, true);
                 return;
             }


vipul674:
@RatLoopz I have submitted the fix. Please let me know if any changes are needed!

PR #2468, opened 2026-06-24
fix(cron): consolidate expiry notifications and fix mark-after-send retry storm in broadcastExpiryAlerts

### LINKED ISSUE #2306: [BUG] : `broadcastExpiryAlerts` Sends O(subscribers × batches) Notifications Per Subscriber, Then Marks All Batches Sent — No Per-Subscriber Guard

### Bug Description

**File:** `apps/api/src/cron/alert-broadcaster.ts` (lines 243–317)

**Description:**

`broadcastExpiryAlerts` loads **all** expiring batches upfront, then for every subscriber page, for every subscriber, sends a notification **for every expiring batch**:

```ts
for (const sub of chunk) {
    for (const batch of expiringBatches) {
        notificationPromises.push(
            sendNotificationToSubscriber(sub, "expiry", { ... })
        );
    }
}
```

### Expected Behavior

With 50 expiring batches and 1,000 subscribers in a page, this fires 50,000 SMS/WhatsApp messages per cron tick. Users receive one alert per expiring batch — not a consolidated summary — which is both spam and cost-prohibitive.

More critically, `expiry_broadcasted` is set to `true` for each batch only **after** all subscriber pages finish (lines 312–316). If the loop breaks mid-way (subscriber fetch error at page 2), the outer catch does not re-enter the per-batch update. All batches remain `expiry_broadcasted = false`. On the next 30-second tick, all subscribers already notified in page 1 get the full set of expiring-batch notifications again. This continues indefinitely until all pages complete without error.

Additionally, `expiringBatches` includes records with `expiry_date` **in the past** (the query uses `.lte("expiry_date", thirtyDaysFromNow)` with no lower bound), so already-expired batches keep getting included until their `expiry_broadcasted` is flipped.

### Screenshots

_No response_

### GSSoC 2026

### Code of Conduct

hariom888:
## Summary
Fixes #2306 — `broadcastExpiryAlerts` had three compounding issues:

1. **O(subscribers × batches) notification fan-out**: for N expiring batches and S subscribers, the cron sent N×S notifications per tick (e.g. 50 batches × 1,000 subscribers = 50,000 sends). Users got one message per expiring batch instead of a consolidated summary.
2. **Retry storm on partial failure**: `expiry_broadcasted` was only set to `true` after *all* subscriber pages completed. Any failure mid-loop (e.g. a subscriber fetch error on page 2) left every batch unmarked, so the full fan-out re-fired to all subscribers — including those already notified in page 1 — on every subsequent 30s tick, indefinitely.
3. **No lower bound on expiry date**: the query used only `.lte('expiry_date', thirtyDaysFromNow)`, so already-expired batches stayed in the result set and kept being notified about until their flag happened to get set.

## Fix
- **Consolidation**: built `sendConsolidatedExpiryNotification`, which sends exactly one message per subscriber listing all currently-expiring batches, reusing the existing per-batch localized `"expiry"` template for each line (joined under one header) rather than introducing new per-language strings.
- **Mark-before-send, per batch**: each batch's `expiry_broadcasted` is now set to `true` individually, *before* subscriber pagination begins at all (stronger than scoping it to a per-batch subscriber loop, since paging is now batch-independent). A batch whose mark-update fails is skipped for this tick's message and naturally retried next tick — it's never lost, and a later subscriber-page failure can no longer cause re-delivery of an already-marked batch.
- **Lower bound**: added `.gte('expiry_date', todayStr)` alongside the existing `.lte(...)`, excluding already-expired batches from the query.

## Testing
New tests in `alertBroadcaster.test.ts`:
- 3 batches × 2 subscribers → exactly 2 sends (consolidated), each containing all 3 batch numbers — not 6.
- Batch marking happens before any subscriber fetch (call-order assertion).
- A batch already marked `expiry_broadcasted = true` from a prior tick is never refetched/resent.
- `.gte('expiry_date', ...)` is asserted on the query, confirming already-expired batches are excluded.
- A batch whose mark-update fails is excluded from the consolidated message while other batches still send correctly.

Full suite: 232/238 passing. The 6 pre-existing failures (`batch.test.ts`, `verify.test.ts`, `originCheck.test.ts`) are unrelated (missing `vitest` dependency, unrelated status-code assertions) and unaffected by this change. `tsc --noEmit` clean.

## Checklist
- [x] Reproduced the O(S×N) fan-out and retry-storm behavior against the described repro steps
- [x] Fix verified via targeted unit tests for consolidation, mark-before-send ordering, and date filtering
- [x] No changes to `broadcastDistrictAlerts` / `broadcastDrugAlerts` (out of scope for this issue)
- [x] Added regression tests for all three sub-issues described


shyam-medh:
@hariom888 how you are handling the checks of PR ?

PR #2470, opened 2026-06-24
fix(api): stop silently aliasing district to city in report submission, breaking district alert thresholds and subscriber matching

### LINKED ISSUE #2307: [BUG] : `district` Field Silently Mapped to `city` in Counterfeit Reports, Breaking District Alert Thresholds and Subscriber Matching

### Bug Description

**File:** `apps/api/src/routes/reports.ts` (lines 108, 130)

**Description:**

The report submission handler explicitly sets `district` to the value of `city`:

```ts
district: data.city,
// ...
district: data.city,   // both in validationPayload and in the DB insert
```

The `createReportSchema` captures `city` (a city name like "Mumbai") but has no `district` field. The API documentation and schema imply these are separate geographic identifiers, but the code conflates them. This has two cascading failures:

1. **District alert threshold is unreachable for multi-city districts.** `PATCH /api/reports/:id/status` counts `verified_fake` reports filtered by `district` (line 247 in reports.ts). If one city submits 4 reports and another city in the same administrative district submits 2, neither city-as-district ever crosses the threshold of 5, even though the real district has 6 verified reports.

2. **Subscriber notification never fires.** `broadcastDistrictAlerts` queries `notification_subscribers` with `.ilike("district", alert.district)` where `alert.district` is a city name. Subscribers who registered with their administrative district (e.g., "Pune District") will never match an alert keyed on city name "Pune". Critical counterfeit drug alerts are silently dropped for registered subscribers.

### Expected Behavior

This is a data integrity issue not just a naming one — city names already exist in the `city` column, so `district` currently contains a redundant copy of `city` rather than the intended administrative district, making geographically-aware alert aggregation inoperable.

### Screenshots

_No response_

### GSSoC 2026

### Code of Conduct

hariom888:
## Summary
Fixes #2307 — `POST /api/reports` always wrote `district: data.city` into both the validation payload and the `counterfeit_reports` insert. `createReportSchema` had no `district` input at all, so `city` silently stood in for `district` everywhere downstream.

## Cascading effects (as described in the issue)
1. **District alert threshold unreachable for multi-city districts**: the threshold count in `PATCH /api/reports/:id/status` (and `admin.controller.ts`'s identical logic) filters by `district`. With `district == city`, reports spread across multiple cities within the same administrative district never aggregate toward the threshold of 5.
2. **Subscriber notification never fires**: `broadcastDistrictAlerts` matches subscribers via `.ilike('district', alert.district)`. Subscribers register with real district names (e.g. "Pune District"); an alert keyed on a city name ("Pune") never matches, so critical alerts were silently dropped.
3. (Found during investigation, not explicitly called out in the issue body but present in code): `reportValidation.service.ts`'s burst/sybil detection treats `city` and `district` as two independent risk signals — collapsing them removed one detection dimension entirely.

## Fix
- Added an optional `district` field to `createReportSchema`, matching the pattern already used in `pharmacies.ts` (which has its own required, independent `district` field).
- When a client provides `district`, it's used as-is — fully independent of `city`.
- When omitted, falls back to `city` for backward compatibility with clients that haven't added a district input yet. This fallback is explicit and documented in code, not a silent permanent aliasing — new clients that send a real district get correct behavior immediately.

## Why no changes to `admin.controller.ts` or `alert-broadcaster.ts`
Both already read the `district` *column* correctly — `.eq("district", data.district)` and `.ilike("district", alert.district)` were never the bug. They simply never received real district data to operate on. Once `reports.ts` writes correct values, both work correctly with zero changes.

## Frontend follow-up
No current report-submission form in `apps/web` in this checkout posts a `city` field at all (the UI either lives elsewhere or isn't yet wired up), so there's no frontend district input to add in this PR. Flagging as a follow-up: once a report form exists, it should collect `district` as a field distinct from `city`, not derive one from the other.

## Testing
- `reports.test.ts`: backward-compat fallback (`city` used as `district` when omitted), explicit `district` used over `city` when both provided, `district` correctly passed to `validateReport`'s burst/sybil detection (not `city`), and schema validation rejecting a too-short explicit `district`.
- `alertBroadcaster.test.ts`: end-to-end test demonstrating an alert keyed on a real administrative district (`"Pune District"`) now matches a subscriber registered under that same district — the exact failure mode described in the issue.
- Full suite: 237/243 passing. The 6 pre-existing failures (missing `vitest` dependency, unrelated status-code assertions) are unaffected by this change.
- `tsc --noEmit` clean.

## Checklist
- [x] Reproduced the city/district conflation and traced its effect through threshold counting and subscriber matching
- [x] Added an explicit, backward-compatible `district` field rather than a breaking schema change
- [x] No changes needed in `admin.controller.ts` / `alert-broadcaster.ts` (verified they were already correct downstream consumers)
- [x] Added regression tests for fallback behavior, explicit-value precedence, validation-service signal independence, and end-to-end subscriber matching


hariom888:
@dipexplorer 

## CI Failures — Pre-existing, Not Caused by This PR

### 1. 🐍 Test Python ML Service (FastAPI) — `tests/test_tts.py`

Two `prune_cache` tests fail (`test_prune_cache_by_file_count`, `test_prune_cache_by_size`) because the TTS cache pruning logic has a race condition with shared `/tmp/tts_cache/` state across concurrent test runs. This file (`routers/tts.py`, `tests/test_tts.py`) is completely unrelated to the alert broadcaster or Redis lock changes in this PR. No file I touched is in the Python ML service.

### 2. 🧪 Build & Test Node.js Apps (Web & API) — Next.js build

Three `Module not found` errors:
- `class-variance-authority` (missing dep in `components/ui/Badge.tsx`)
- `@opentelemetry/sdk-trace-web` / `auto-instrumentations-web` / `context-zone` (missing deps in `lib/tracing.ts`)

Plus a duplicate parameter name parse error in `app/[locale]/expiry-tracker/components/ExpiryForm.tsx` (`isSubmitting`
declared twice in the function signature).

None of these files exist in my branch. These are missing `package.json` dependencies and a pre-existing syntax bug in the `web` app — both present on `main` before my branch was created.

PR #2474, opened 2026-06-24
fix(alert-broadcaster): Redis distributed lock prevents duplicate SMS/WhatsApp on horizontal scale

### LINKED ISSUE #2308: [BUG] : Concurrent Multi-Instance Alert Broadcaster Has No Distributed Lock — Duplicate SMS/WhatsApp Sent to All Subscribers

### Bug Description

**File:** `apps/api/src/cron/alert-broadcaster.ts` (lines 6–8, `startAlertBroadcaster`), `apps/api/src/index.ts` (line 18)

**Description:**

`startAlertBroadcaster` uses a module-level `intervalId` guard to prevent double-registration within a single process:

```ts
let intervalId: NodeJS.Timeout | null = null;
// ...
if (intervalId) {
    logger.warn("Alert broadcaster is already running.");
    return;
}
```

This guard only prevents re-entry within the **same Node.js process**. In any horizontally scaled deployment (multiple Docker containers, Kubernetes pods, or PM2 cluster mode), every process runs its own `startAlertBroadcaster`. Each process independently:
1. Fetches the same unbroadcasted alerts (`.eq("broadcasted", false)`).
2. Iterates over all subscriber pages.
3. Sends the full notification batch.
4. Marks `broadcasted = true`.

Between step 1 and step 4 across two concurrent processes, both read `broadcasted = false` (no SELECT FOR UPDATE or optimistic lock), both dispatch the full subscriber set, and both eventually flip the flag. Every subscriber receives **at minimum 2×** the notifications per alert cycle. At 1,000 subscribers per alert and 3 alert types, this is 6,000 duplicate medical safety messages (SMS/WhatsApp) per 30-second tick until one process completes and sets the flag.

### Expected Behavior

There is no Redis-based distributed lock, no Postgres advisory lock, and no database-level row claim (e.g., `UPDATE ... SET claimed_at = now() WHERE claimed_at IS NULL RETURNING *`) in any of the three broadcaster functions.

### Screenshots

_No response_

### GSSoC 2026

### Code of Conduct

hariom888:
## Summary

Fixes #2308.

`startAlertBroadcaster` uses a process-local `intervalId` guard that is invisible to other pods. In any horizontally scaled deployment every pod independently fetches `broadcasted = false`, sends the full subscriber batch, then flips the flag — causing **N-pod × subscriber-count** duplicate medical safety messages per 30-second tick.

## Root Cause

`checkAndBroadcastAll` had no distributed mutual exclusion between the `SELECT broadcasted = false` read and the `UPDATE broadcasted = true` write. Any two pods racing the same tick both see `false`, both send, both mark `true`.

## Fix

Wrap `checkAndBroadcastAll` in a **Redis `SET NX PX` distributed lock** using the already-available `redisClient` from `apps/api/src/utils/redis.ts`.

- Lock key: `alert-broadcaster:lock`  
- Lock value: `${HOSTNAME}:${pid}` — unique per pod, used for safe release  
- TTL: `25 000 ms` (intentionally under the 30-second interval to prevent stale locks blocking the next tick if a pod crashes after acquiring)  
- Release: Lua compare-and-delete — only the owning pod can release its lock, preventing accidental release of a lock re-acquired by another pod after TTL expiry  
- Graceful fallback: if Redis is not connected (local dev, test env), a warning is logged and the broadcaster runs unlocked — same behaviour as before

## Files Changed

| File | Change |
|------|--------|
| `apps/api/src/cron/alert-broadcaster.ts` | Import `redisClient`; add `LOCK_KEY/TTL/VALUE` constants; add `acquireLock`/`releaseLock` helpers; wrap `checkAndBroadcastAll` in lock/finally |
| `apps/api/src/cron/alert-broadcaster.test.ts` | New unit tests: lock acquired + released on normal run; second invocation skips when lock held; lock released in finally; fallback when Redis disconnected |

## Testing

- Existing `apps/api/src/services/notifications.test.ts` — unaffected (no broadcaster import)
- New `alert-broadcaster.test.ts` covers all four lock paths
- Manual: start two API instances pointing at the same Redis; confirm only one pod logs broadcast activity per tick and subscribers receive exactly one message

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [x] Follows project code style
- [x] Tests added for new behaviour
- [x] No breaking changes to existing tests
- [x] Targets `main` branch

### DIFF
--- a/apps/api/src/cron/alert-broadcaster.test.ts
+++ b/apps/api/src/cron/alert-broadcaster.test.ts
@@ -1,0 +1,102 @@
+// Fix: Mock WebSocket for Node.js versions < 22 to prevent Supabase Realtime client crash during test imports
+if (typeof globalThis.WebSocket === "undefined") {
+    globalThis.WebSocket = class {} as any;
+}
+
+import assert from "node:assert/strict";
+import { describe, it, beforeEach, mock } from "node:test";
+
+// ── Shared Redis mock state ────────────────────────────────────────────────
+const lockStore = new Map<string, string>();
+
+const mockRedisClient = {
+    isOpen: true,
+    set: mock.fn(async (key: string, value: string, opts?: { NX?: boolean; PX?: number }) => {
+        if (opts?.NX && lockStore.has(key)) return null; // NX: only set if not exists
+        lockStore.set(key, value);
+        return "OK";
+    }),
+    eval: mock.fn(async (_script: string, { keys, arguments: args }: { keys: string[]; arguments: string[] }) => {
+        const [key] = keys;
+        const [expected] = args;
+        if (lockStore.get(key) === expected) {
+            lockStore.delete(key);
+            return 1;
+        }
+        return 0;
+    }),
+};
+
+// ── Module mocks ───────────────────────────────────────────────────────────
+mock.module("../utils/redis", () => ({ redisClient: mockRedisClient }));
+
+mock.module("../db/client", () => ({
+    supabase: {
+        from: () => ({
+            select: () => ({ eq: () => ({ data: [], error: null }) }),
+            update: () => ({ eq: () => ({ error: null }) }),
+        }),
+    },
+    dbConfig: { isSupabaseOffline: false },
+}));
+
+mock.module("../services/sms-service", () => ({ smsService: { send: async () => true } }));
+mock.module("../services/whatsapp-service", () => ({ whatsappService: { send: async () => true } }));
+mock.module("../utils/logger", () => ({
+    default: { info: () => {}, warn: () => {}, error: () => {}, debug: () => {} },
+}));
+
+import { checkAndBroadcastAll } from "./alert-broadcaster";
+
+describe("checkAndBroadcastAll — distributed lock", () => {
+    beforeEach(() => {
+        lockStore.clear();
+        mockRedisClient.set.mock.resetCalls();
+        mockRedisClient.eval.mock.resetCalls();
+    });
+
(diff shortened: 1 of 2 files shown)


hariom888:
@dipexplorer 

## CI Failures — Pre-existing, Not Caused by This PR

### 1. 🐍 Test Python ML Service (FastAPI) — `tests/test_tts.py`

Two `prune_cache` tests fail (`test_prune_cache_by_file_count`, `test_prune_cache_by_size`) because the TTS cache pruning logic has a race condition with shared `/tmp/tts_cache/` state across concurrent test runs. This file (`routers/tts.py`, `tests/test_tts.py`) is completely unrelated to the alert broadcaster or Redis lock changes in this PR. No file I touched is in the Python ML service.

### 2. 🧪 Build & Test Node.js Apps (Web & API) — Next.js build

Three `Module not found` errors:
- `class-variance-authority` (missing dep in `components/ui/Badge.tsx`)
- `@opentelemetry/sdk-trace-web` / `auto-instrumentations-web` / `context-zone` (missing deps in `lib/tracing.ts`)

Plus a duplicate parameter name parse error in `app/[locale]/expiry-tracker/components/ExpiryForm.tsx` (`isSubmitting`
declared twice in the function signature).

None of these files exist in my branch. These are missing `package.json` dependencies and a pre-existing syntax bug in the `web` app — both present on `main` before my branch was created.

---

This PR only modifies:
- `apps/api/src/cron/alert-broadcaster.ts`
- `apps/api/src/cron/alert-broadcaster.test.ts`

PR #2753, opened 2026-06-28
[BUG] Fix cache key mismatch: warmCache() batch-only keys vs drug lookup composite keys — 100% cache miss rate

### LINKED ISSUE #2750: [BUG] Cache key mismatch: warmCache() uses batch-only keys while drug lookups use composite keys — 100% cache miss rate

### Bug Description

The cache warming function `warmCache()` in `apps/api/src/services/cache.service.ts` stores entries using a pure batch-number key (`drug:batch:BN2024001`), but drug lookups in `apps/api/src/services/drugLookup.service.ts` always use a three-part composite key (`BN2024001||BrandName` or `BN2024001|barcode123|BrandName`). Because `getCachedDrug()` constructs `drug:batch:${batchNumber}` from whatever key is passed in, the two key formats never align — causing a **100% cache miss rate** for pre-warmed data.

**File 1:** `apps/api/src/services/cache.service.ts:76`
```typescript
// warmCache() stores with batch-only key:
const cacheKey = `${KEY_PREFIXES.DRUG_CACHE}${med.batch_number}`;  // key = drug:batch:BN2024001
```

**File 2:** `apps/api/src/services/drugLookup.service.ts:27,31`
```typescript
// lookupDrugByBatch() creates a three-part composite key:
const cacheKey = `${batchNumber}|${identifier.barcode_id || ""}|${identifier.brand_name || ""}`;
// e.g., "BN2024001|barcode123|BrandName" or "BN2024001||BrandName"

// Then calls getCachedDrug() with the composite key:
const cachedDrug = await getCachedDrug(cacheKey);
```

**File 3:** `apps/api/src/services/cache.service.ts:141-144`
```typescript
// getCachedDrug() uses parameter directly as cache key:
export async function getCachedDrug(batchNumber: string): Promise<any | null> {
    const cacheKey = `${KEY_PREFIXES.DRUG_CACHE}${batchNumber}`;
    // cacheKey = drug:batch:BN2024001|barcode123|BrandName
```

**Result:** `warmCache()` writes to `drug:batch:BN2024001`, but `getCachedDrug()` reads from `drug:batch:BN2024001||BrandName` — entirely different keys. All 47 batch entries pre-seeded during cache warming on server startup are wasted.

### Expected Behavior

Pre-warmed cache entries should be discoverable by drug lookups. The key format between `warmCache()` and `lookupDrugByBatch()` should be consistent, or `lookupDrugByBatch()` should fall back to a batch-only cache check when the composite-key lookup misses.

### Impact

- **Cache warming is completely ineffective** — 100% of pre-warmed drug entries never match any lookup.
- **Every lookup starts as a cache miss**, hitting Postgres every first time (even for drugs that were pre-warmed).
- Startup performance benefit from `warmCache()` is zero.
- Redis memory stores duplicate data: one batch-only entry (never read) plus one composite entry per unique barcode/brand combination.

### Suggested Fix

Either:

**Option A (preferred):** Normalize `warmCache()` to store using a consistent composite key format, or store the same drug data under multiple key formats so both batch-only and composite lookups match.

**Option B:** After a composite-key cache miss in `getCachedDrug()`, fall back to checking a batch-only key before hitting the database.

**Option C:** Make `lookupDrugByBatch()` check the batch-only cache key first, then only use the composite key if a brand/barcode filter is needed.

### GSSoC 2026

vipul674:

## 📋 PR Summary & Link

- **Closes #2750**
- **Summary:** `warmCache()` stored entries under batch-only keys (`drug:batch:BN2024001`) but `lookupDrugByBatch()` constructed composite keys (`BN2024001||BrandName`), causing 100% cache misses. Fix: check batch-only key first, then composite key, then DB. When caching from DB, store under both formats so pre-warmed data is discoverable.

## 🏷️ PR Type

- [x] 🐛 `type: bug`

## ✅ Checklist

- [x] My PR has a linked issue (`Closes #2750`)

### DIFF
--- a/apps/api/src/services/drugLookup.service.ts
+++ b/apps/api/src/services/drugLookup.service.ts
@@ -24,21 +24,38 @@ export async function lookupDrugByBatch(
         throw new Error("Either brand_name or barcode_id must be provided alongside batch_number");
     }
 
-    const cacheKey = `${batchNumber}|${identifier.barcode_id || ""}|${identifier.brand_name || ""}`;
+    const compositeKey = `${batchNumber}|${identifier.barcode_id || ""}|${identifier.brand_name || ""}`;
 
-    // 1. Try to fetch from Redis cache
+    // 1. Try batch-only cache key first (matches warmCache() format)
     try {
-        const cachedDrug = await getCachedDrug(cacheKey);
+        const batchOnlyDrug = await getCachedDrug(batchNumber);
+        if (batchOnlyDrug) {
+            const matchesIdentifier =
+                (!identifier.barcode_id || batchOnlyDrug.barcode_id === identifier.barcode_id) &&
+                (!identifier.brand_name || batchOnlyDrug.brand_name === identifier.brand_name);
+            if (matchesIdentifier) {
+                logger.info(`Cache HIT (batch-only key) for drug batch: ${batchNumber}`);
+                await incrementHitCount(batchOnlyDrug.id, batchOnlyDrug.brand_name || batchOnlyDrug.generic_name);
+                return batchOnlyDrug;
+            }
+        }
+    } catch (err) {
+        logger.error(`Error checking batch-only cache for batch: ${batchNumber}`, err);
+    }
+
+    // 2. Try composite cache key
+    try {
+        const cachedDrug = await getCachedDrug(compositeKey);
         if (cachedDrug) {
-            logger.info(`Cache HIT for drug batch: ${cacheKey}`);
+            logger.info(`Cache HIT (composite key) for drug batch: ${compositeKey}`);
             return cachedDrug;
         }
     } catch (err) {
-        logger.error(`Error checking cache for batch: ${cacheKey}`, err);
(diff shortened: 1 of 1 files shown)


vipul674:
@RatLoopz PR #2753 ready for review.

## Fix Report — Issue #2750

**Problem:** `warmCache()` stored entries under batch-only keys (`drug:batch:BN2024001`) but `lookupDrugByBatch()` constructed composite keys (`BN2024001||BrandName`), causing 100% cache miss rate — cache warming was completely useless.

**Fix (Option B approach):**
1. `lookupDrugByBatch()` now checks the batch-only cache key first (matches `warmCache()` format)
2. Falls back to composite key if batch-only miss
3. On DB fetch, stores under **both** key formats so warm cache and composite lookups both succeed

**File changed:** `apps/api/src/services/drugLookup.service.ts`
**Type:** Bug fix (cache correctness)

vipul674:
@dipexplorer could you please review and merge this PR? Fixes 100% cache miss rate between warmCache() and drug lookups.

PR #3041, opened 2026-07-03
fix(ml): guard audio.content_type against None in voice_verify router

### LINKED ISSUE #3037: [BUG] FastAPI UploadFile.content_type can be None, causing TypeError in voice_verify router

## Bug Description
**Summary:** The `/voice/verify` endpoint in `apps/ml/routers/voice_verify.py` checks `audio.content_type not in [...]` without guarding against `None`. FastAPI's `UploadFile.content_type` can be `None` when the client omits the Content-Type header in the multipart form-data request. This causes a `TypeError: argument of type 'NoneType' is not iterable`.

**Steps to reproduce:**
1. Send a POST request to `/voice/verify` with a multipart form containing an audio file but without setting the `Content-Type` header on the file part.
2. Observe that the server crashes with a 500 error instead of returning a 400 error.

**Root cause:** Line 44 in `apps/ml/routers/voice_verify.py`:
```python
if audio.content_type not in ["audio/webm", "audio/wav", "audio/ogg", "audio/mp4", "audio/mpeg"]:
```
When `audio.content_type` is `None`, Python's `not in` check raises `TypeError` instead of returning `True`.

Note that `audio.size` at line 40 is correctly guarded:
```python
if audio.size is None or audio.size == 0:
```

The same guard is missing for `audio.content_type`. Compare with `apps/ml/routers/asr.py` which correctly handles nullable `content_type` via its `normalize_content_type()` helper.

## Expected Behavior
The endpoint should return a 400 error with "Invalid audio format" when `content_type` is missing, rather than crashing with a 500 error.

## GSSoC 2026

## Code of Conduct

vipul674:

## 📋 PR Summary & Link

- **Closes #3037**
- **Summary:** Guard `audio.content_type` against `None` before membership test in `voice_verify.py`. When `content_type` is `None` (client omits Content-Type header), `None not in [...]` raises `TypeError`. Fix adds `not audio.content_type` guard to return proper 400 error.

## 🏷️ PR Type

- [x] 🐛 `type: bug`

## ✅ Checklist

- [x] My PR has a linked issue (`Closes #3037`)

### DIFF
--- a/apps/ml/routers/voice_verify.py
+++ b/apps/ml/routers/voice_verify.py
@@ -41,7 +41,7 @@ async def verify_medicine_voice(audio: UploadFile = File(...)):
     detects language, and verifies the medicine via LangChain + CDSCO.
     """
     # Validate file type
-    if audio.content_type not in ["audio/webm", "audio/wav", "audio/ogg", "audio/mp4", "audio/mpeg"]:
+    if not audio.content_type or audio.content_type not in ["audio/webm", "audio/wav", "audio/ogg", "audio/mp4", "audio/mpeg"]:
         raise HTTPException(status_code=400, detail="Unsupported audio format. Use webm, wav, ogg, or mp4.")
 
     # Save to temp file for Whisper


vipul674:
@dipexplorer PR ready for review.

**Change:** Guard `audio.content_type` against `None` before `not in` test in `apps/ml/routers/voice_verify.py:44`. Returns 400 instead of 500 crash when client omits Content-Type header.

Closes #3037

PR #3410, opened 2026-07-09
Fix: close idempotency middleware TOCTOU race allowing duplicate scan submissions

### LINKED ISSUE #3393: [BUG] : Idempotency middleware has a TOCTOU race allowing duplicate scan submissions under concurrent retries

### Bug Description

`idempotencyMiddleware` (`apps/api/src/middleware/idempotency.ts`) only performs a read (Redis GET, then Supabase SELECT on `submission_idempotency`) before calling `next()`. The actual write — both the Redis SET and the Supabase INSERT into `submission_idempotency` — happens at the very end of the `/submit` handler in `apps/api/src/routes/scan.ts` (lines ~977-985), after file upload validation, conflict resolution (`resolveConflict`), and part-status upserts have all completed.

Two requests carrying the same `Idempotency-Key` sent close together (a normal client retry pattern on flaky mobile networks — this endpoint is explicitly for offline sync) will both pass the middleware's existence check, both execute the full submission pipeline, and both attempt the final insert into `submission_idempotency` (`idempotency_key` is PRIMARY KEY). The second insert's error result is discarded — `await supabase.from("submission_idempotency").insert(...)` is never checked — so it fails silently while the handler still returns 200 with a full result. `resolveConflict` can run twice concurrently for the same logical submission, and the durable idempotency record for the "losing" request is never persisted, so a later retry with the same key won't find it and will reprocess the submission a third time.

### Expected Behavior

A durable idempotency reservation should be made atomically before any processing begins (e.g. an INSERT relying on the PRIMARY KEY conflict inside the middleware itself), so concurrent requests with the same key short-circuit before hitting the handler, and any insert failure is logged rather than silently discarded.

### Screenshots

_No response_

### GSSoC 2026

### Code of Conduct

Srejoye:
## Summary
Closes #3393. `idempotencyMiddleware` only performed reads (Redis GET, then a Supabase SELECT) before calling `next()`. The durable write into `submission_idempotency` happened at the end of the `/submit` handler, *after* upload validation, `resolveConflict`, and part-status upserts — and its result was never checked. Two concurrent requests carrying the same `Idempotency-Key` (a normal retry pattern on flaky mobile networks) could both pass the middleware, both run the full pipeline, and the loser's insert would fail silently on the primary-key conflict.

## Fix
The durable reservation now happens atomically in the middleware, before any processing begins:

1. Middleware attempts `INSERT INTO submission_idempotency (idempotency_key, scan_id: null)`.
   - **Insert succeeds** → this request owns the key → proceeds to the handler.
   - **Insert fails, `error.code === '23505'`, `scan_id` present** → a prior request already finished → return its cached result (`200`).
   - **Insert fails, `error.code === '23505'`, `scan_id` still `null`** → a prior request for the same key is currently in-flight → short-circuit with `409` instead of letting this request also run the pipeline.
   - **Any other DB error** → fail closed (`500`), logged instead of discarded.
2. The `/submit` handler no longer inserts a second time. Once `resolveConflict` and the part upserts complete, it `UPDATE`s the pre-reserved row's `scan_id` and logs (rather than swallows) any error.
3. On a genuine handler failure, the reservation row is deleted so a crashed/failed attempt doesn't permanently lock out legitimate retries of the same `Idempotency-Key`.

This relies on the existing `idempotency_key TEXT PRIMARY KEY` constraint on `submission_idempotency`, so the race is closed at the database level, not just in application logic.

## Why this is safe
- `scan_id` on `submission_idempotency` is nullable, so reserving with `scan_id: null` before it's known is a valid write.
- `resolveConflict`, part-status upserts, and the final Redis cache write are unchanged.
- Failure-path cleanup (`DELETE` in the `catch` block) is best-effort and wrapped so a secondary failure there can't mask the original error response.

## Testing
`apps/api/tests/scans.sync.test.ts` rewritten to cover:
- Redis cache hit (unchanged behavior).
- Happy path: reservation succeeds, submission completes, reservation row updated with the resolved `scan_id` (not a second insert).
- Concurrent duplicate still in-flight → `409` before the handler runs (`scan_submission_parts` upsert never called).
- Concurrent duplicate that already completed → `200` with the durable cached result and parts status.

### DIFF
--- a/apps/api/src/middleware/idempotency.ts
+++ b/apps/api/src/middleware/idempotency.ts
@@ -1,6 +1,7 @@
 import { Request, Response, NextFunction } from "express";
 import { redisClient } from "../utils/redis";
 import { supabase } from "../db/client";
+import logger from "../utils/logger";
 
 export const idempotencyMiddleware = async (req: Request, res: Response, next: NextFunction) => {
     const rawKey = req.headers["idempotency-key"] as string;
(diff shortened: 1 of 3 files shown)


Srejoye:
The files in the failing checks are unrelated to my changes.

PR #3414, opened 2026-07-09
fix(api): reject unparseable clientUpdatedAt instead of silently discarding updates

### LINKED ISSUE #3394: [BUG] : Offline-sync conflict resolution silently discards updates when clientUpdatedAt is missing or malformed

### Bug Description

`resolveConflict` in `apps/api/src/utils/conflictResolution.ts` (lines 59-85) decides whether to apply an incoming update with:
`new Date(Number(input.clientUpdatedAt)) > new Date(existing.client_updated_at || 0)`

`clientUpdatedAt` is an optional string field in `submitSchema` (`apps/api/src/routes/scan.ts`). If the client omits it, `Number(undefined)` is `NaN`, `new Date(NaN)` is an Invalid Date, and any comparison against it evaluates to `false`. The same happens for a non-numeric string. In both cases `incomingIsNewer` is always false, so the update branch is never taken — the new metadata is discarded, and the write is logged to `scan_conflict_log` with resolution `"rejected_stale"`, even when there's no real conflict (e.g. a single-device write, or the first sync of a scanId). This silently drops legitimate updates for any client that doesn't set `clientUpdatedAt`, which isn't currently enforced by validation.

### Expected Behavior

`clientUpdatedAt` should be required or validated as a numeric string in `submitSchema`, and `resolveConflict` should explicitly reject (400) requests where the timestamp can't be parsed, rather than silently logging them as `"rejected_stale"` indistinguishably from a genuine conflict.

### Screenshots

_No response_

### GSSoC 2026

### Code of Conduct

Srejoye:
## Summary
Fixes #3394. `resolveConflict` (`apps/api/src/utils/conflictResolution.ts`) compared `new Date(Number(input.clientUpdatedAt))` against the existing row's timestamp to decide whether an incoming write was newer. When `clientUpdatedAt` was missing or a non-numeric string, `Number()` returned `NaN`, so the comparison always evaluated to `false` — the update was silently discarded and logged to `scan_conflict_log` with resolution `"rejected_stale"`, indistinguishable from a genuine conflict.

## Changes
- **`apps/api/src/routes/scan.ts`**: `clientUpdatedAt` in `submitSchema` is now required (previously `.optional()`) and validated with a regex to be a numeric (epoch-ms) string. Malformed requests are rejected with 400 before conflict resolution ever runs.
- **`apps/api/src/utils/conflictResolution.ts`**: added `InvalidClientTimestampError` and a `parseClientUpdatedAt` helper that validates the timestamp up front and throws before any DB read/write — defense in depth in case `resolveConflict` is ever called from another entry point that bypasses `submitSchema`.
- **`apps/api/src/routes/scan.ts`** `/submit` handler: catches `InvalidClientTimestampError` explicitly and returns 400, instead of falling through to the generic 500 handler.


Srejoye:
The failing tests are due to errors in files which were previously present on main, nothing due to my PR.

PR #3428, opened 2026-07-09
fix(api): add distributed Redis lock to medicine expiry cron

### LINKED ISSUE #3395: [BUG] : Medicine expiry-check cron has no distributed lock, causing duplicate SMS/push notifications when the API is scaled horizontally

### Bug Description

`alert-broadcaster.ts` implements a Redis NX distributed lock specifically because multiple API instances can run the same in-process cron logic (see `docker-compose.prod.yml`). `initExpiryCron` in `apps/api/src/cron/expiry-check.ts` registers its own `node-cron` job directly with no locking at all.

If more than one API instance is running, every instance's scheduler fires the midnight job simultaneously. Each instance independently queries `tracked_medicines WHERE notified_Xd = false`, gets the same rows (none of them flagged yet), and sends push/SMS notifications to every affected user — once per instance. The flag update (`notified_${days}d = true`) only happens in a single bulk update at the end of each window's loop, well after notifications have already gone out, so it provides no protection against concurrent instances.

### Expected Behavior

The cron callback body should be wrapped in the same `acquireLock`/`releaseLock` pattern already implemented in `alert-broadcaster.ts` (ideally extracted into a shared utility), keyed uniquely per job (e.g. `"expiry-check:lock"`).

### Screenshots

_No response_

### GSSoC 2026

### Code of Conduct

Srejoye:
## Summary

Fixes #3395.

The medicine expiry cron previously executed independently on every running API instance. In a horizontally scaled deployment, each scheduler queried the same unnotified rows and sent duplicate SMS and push notifications before the notification flags were updated.

This PR wraps the cron execution with the same Redis distributed lock pattern already used by `alert-broadcaster.ts`.

## Changes

- Added Redis distributed lock to `expiry-check.ts`
- Added `acquireLock()` and `releaseLock()` helpers
- Introduced dedicated lock key:
  - `expiry-check:lock`
- Wrapped the scheduled job so only one API instance performs expiry processing
- Ensured the lock is always released via `finally`

## Result

Only one API instance executes the midnight expiry job, preventing duplicate SMS and push notifications in horizontally scaled deployments.

### DIFF
--- a/apps/api/src/cron/expiry-check.ts
+++ b/apps/api/src/cron/expiry-check.ts
@@ -2,6 +2,7 @@ const cron = require("node-cron");
 import webPush from "web-push";
 import { supabase } from "../db/client";
 import logger from "../utils/logger";
+import { redisClient } from "../utils/redis";
 import { smsService } from "../services/sms-service";
 import {
     listPushSubscriptions,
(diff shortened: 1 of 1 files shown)


Srejoye:
Failed tests are unrelated to this pr.

PR #3587, opened 2026-07-14
fix(api): apply CSRF middleware consistently across environments

### LINKED ISSUE #3552: [BUG] [SECURITY] Conditional CSRF protection masks Missing CSRF middleware in Dev (CodeQL Alert)
## Problem
CodeQL has flagged a "Missing CSRF middleware" vulnerability in `apps/api/src/app.ts` (Alert 136).

Currently, CSRF protection is completely bypassed if the app is in the test or development environment:
```typescript
const isTestOrDev = process.env.NODE_ENV === "test" || process.env.NODE_ENV === "development";
if (!isTestOrDev) {
    app.use(doubleCsrfProtection);
}
```
While bypassing this in local dev environments is convenient, it causes static analysis tools like CodeQL to detect paths that lack CSRF middleware. Furthermore, completely disabling security features in development hides potential integration bugs until the app is pushed to production, where CSRF failures will suddenly appear.

## Files to Touch
- `apps/api/src/app.ts` (Lines 127-131)

## How to Fix
Instead of completely disabling the `doubleCsrfProtection` middleware based on the environment, you should apply it consistently. For automated testing, use a mock token or override the validation function logic within the test runner scope, rather than globally removing the middleware in `app.ts`. Or if it must be bypassed, restrict the bypass to `NODE_ENV === "test"` only and apply it in `"development"`.

## Acceptance Criteria
- [ ] CSRF middleware is applied more consistently to resolve the CodeQL alert.
- [ ] Dev environments mirror production CSRF behaviors to avoid late-stage integration failures.

shashank03-dev:

## 📋 PR Summary & Link

- **Closes #3552**
- **Summary:**

The old code only mounted `doubleCsrfProtection` when `NODE_ENV` was neither `test` nor `development`. That left route paths with no CSRF middleware on them at all, which is what CodeQL flagged in Alert 136, and it meant dev ran with protection switched off.

The middleware is now mounted unconditionally. What used to be an environment check around `app.use()` became a `skipCsrfProtection` callback on the `doubleCsrf()` config, and that callback only returns true for `NODE_ENV === "test"`. So the middleware is always in the chain, which is what the CodeQL query looks for, and development validates tokens the same way production does.

Two things in here that reviewers should look at closely.

Dev needs a secret now. Since the middleware actually runs in development, `getSecret()` gets called there, and it used to throw when `CSRF_SECRET` was unset. Rather than make every dev request 500, it now mints a random 32-byte secret once per process and logs a warning. It comes from `crypto.randomBytes`, not a hardcoded string, so it is never a predictable credential, and it does not survive a restart. Production cannot reach that branch: `app.ts` already calls `process.exit(1)` at startup if `CSRF_SECRET` is missing outside dev and test.

The test skip is captured at module load, not per request. Some existing suites flip `NODE_ENV` mid-run to exercise production branches. If the skip were evaluated per request, those suites would start failing CSRF checks for reasons that have nothing to do with what they are testing. Reading it once at load keeps the same timing the old env gate had.

No frontend changes were needed. `apps/web/lib/api.ts` already fetches `/api/csrf-token`, sends `x-csrf-token`, and refreshes the token on a 403.

I booted the API with `NODE_ENV=development` and sent real requests, first on `main` and then on this branch.

Before, on `main` @ `bd718cf1`. A forged cross-site POST carrying no CSRF token is accepted and passes straight through to the route's own validation:

```
### BEFORE (main @ bd718cf1, NODE_ENV=development) — forged POST, no CSRF token:
   -> HTTP 400   (not 403 => CSRF middleware absent, request accepted)
```

After, on this branch. The forged request is rejected, the real double-submit flow gets through, and a tampered token is rejected:

```
### 1. Forged POST — no CSRF token (attacker's cross-site request)
   -> HTTP 403

### 2. Legit flow — fetch token from /api/csrf-token, then POST with it
   token: b78904ff3a97b74547e9010582993ed7...
   -> HTTP 400  (NOT 403 — CSRF passed; 400 is the route's own body validation)

### 3. Tampered token
   -> HTTP 403
```

`apps/api/tests/csrfProtection.test.ts` locks this in. It loads a fresh app under `NODE_ENV=development` so the skip is inactive, then asserts both directions of the double-submit check. It aims at a path with no route behind it, so what gets asserted is the global middleware and not some individual route's auth or body validation:

```
$ npx jest tests/csrfProtection.test.ts
Test Suites: 1 passed, 1 total
Tests:       2 passed, 2 total
```

On regressions: the full API suite comes out the same on this branch as on `main`. 536 passing, and the 7 failures across 5 suites (`requestId`, `adminReportStatus`, `map`, `cache.service`, `abha.service`) reproduce identically on `main` @ `bd718cf1` whether this branch is checked out or not. They are pre-existing and I have not touched them.

## 🏷️ PR Type

- [ ] 🐛 `type: bug`
- [ ] ✨ `type: feature`
- [ ] 📖 `type: docs`
- [ ] 🧪 `type: testing`
- [x] 🔒 `type: security`
- [ ] ⚡ `type: performance`
- [ ] 🎨 `type: design`
- [ ] ♻️ `type: refactor`
- [ ] 🛠️ `type: devops`
- [ ] ♿ `type: accessibility`

## ✅ Checklist

- [x] My PR has a linked issue (`Closes #3552`)

### DIFF
--- a/apps/api/src/app.ts
+++ b/apps/api/src/app.ts
@@ -98,14 +98,28 @@ app.use(cors(createCorsOptions()));
 // csrf-csrf is recognized by CodeQL as a valid CSRF defense unlike custom header checks.
 const ANON_SESSION_COOKIE = "csrf_anon_id";
 
+// Decided once at startup so it stays stable even if a test mutates NODE_ENV.
+const SKIP_CSRF_VALIDATION = process.env.NODE_ENV === "test";
+
+// Ephemeral fallback secret for local dev when CSRF_SECRET is unset (see getSecret).
+let devCsrfSecret: string | undefined;
+
 const { doubleCsrfProtection, generateCsrfToken: generateToken } = doubleCsrf({
     getSecret: () => {
         const secret = process.env.CSRF_SECRET;
-        if (!secret) {
-            logger.error("CSRF_SECRET environment variable is not set");
-            throw new Error("CSRF_SECRET environment variable is required");
+        if (secret) return secret;
+        // Production never reaches here: startup exits when CSRF_SECRET is missing
+        // outside dev/test. Now that the middleware runs in development too, mint a
+        // random per-process secret so local dev works without extra setup instead
+        // of 500-ing every request. Generated (not hardcoded) so it is never a
+        // predictable credential; it simply won't survive a restart.
+        if (!devCsrfSecret) {
+            devCsrfSecret = crypto.randomBytes(32).toString("hex");
+            logger.warn(
+                "CSRF_SECRET is not set — using an ephemeral per-process dev secret. Set CSRF_SECRET (see .env.example) outside local development."
+            );
         }
-        return secret;
+        return devCsrfSecret;
     },
     getSessionIdentifier: (req: Request) => {
         if (req.cookies?.access_token) {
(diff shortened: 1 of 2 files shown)


shashank03-dev:
Quick note on CI, since two checks went red on the first run.

The Playwright job was a flake. It failed on `expiry-tracker.spec.ts` with `Target page, context or browser has been closed`, which is the browser dying rather than anything the app did. I reran the workflow on the same commit (`d5fc434c`, no changes) and it passed: run 29331616799. Worth saying explicitly that the e2e workflow never builds or starts `apps/api` at all, it builds the shared packages and the web app and Playwright's webServer only boots Next, so the CSRF middleware in this PR is not loaded during that job. The offline/IndexedDB specs have flaked on unrelated branches too, including a docs-only one.

`Build & Test Node.js Apps` is still failing, but it fails the same way on main. It dies on `Cannot find module '@sahidawa/validators' from 'src/routes/reports.ts'`, and main's own run at bd718cf1 (29313599749) has the identical error. It also fails on my Python-only PR #3588, which does not touch any TS. So it predates this branch and I have left it alone rather than widening the diff. Happy to open a separate issue for it if that is useful.

PR #3586, opened 2026-07-14
fix(a11y): add ARIA live regions and keyboard focus management to InteractiveOverlays

yogita-mehta:
Closes #3272

## What this does
Adds keyboard focus management and dialog semantics to the Chatbot and CommandPalette overlays extracted in `InteractiveOverlays.tsx`, so keyboard and screen reader users aren't left interacting with the hidden background page while an overlay is open.

## Approach
The repo already had a fully-implemented `useFocusTrap` hook (with its own test suite) that wasn't wired into either overlay yet. Rather than reimplementing focus-trapping from scratch, this PR wires it in and adds the missing dialog semantics on top:

**Chatbot.tsx**
- Wired in the existing `useFocusTrap(chatbotRef, isOpen)` hook
- Added `role="dialog"`, `aria-modal`, `aria-hidden`, and `aria-label` to the panel
- Added `inert={!isOpen}` — since this panel stays mounted and only fades via CSS transitions (rather than unmounting), `inert` is needed to remove it from the tab order and screen reader tree while closed; `aria-hidden` alone isn't sufficient for a keyboard-only user
- Added a visually-hidden `aria-live="polite"` region that announces when the assistant opens

**CommandPalette.tsx**
- Wired in `useFocusTrap(containerRef, isOpen)`
- Added `role="dialog"`, `aria-modal="true"`, and `aria-label`
- No separate live region needed here since this component fully unmounts on close (`if (!isOpen) return null`), so its `role="dialog"` and label are picked up naturally when it mounts

## Testing
- Manually verified keyboard-only navigation: Tab/Shift+Tab stay contained within both overlays while open, wrapping correctly at the first/last focusable element
- Verified closing each overlay returns focus appropriately (Chatbot: back to the trigger bubble via the hook's built-in focus restoration, which already has test coverage in `useFocusTrap.test.tsx`)
- Confirmed no visual regressions — the added attributes are behavioral only
- Ran `npm run lint` and `npx tsc --noEmit` — no new errors introduced (pre-existing `@sahidawa/*` workspace build and jest tsconfig deprecation issues were confirmed present on clean `main` via `git stash`, unrelated to this change)

PR #3590, opened 2026-07-14
fix: add crisp neutral outline ring to Back to Top button for light mode visibility

### LINKED ISSUE #3051: [BUG] Back to Top button is not visible in light mode

### Bug Description

The Back to Top button is difficult to see when the application is in light mode. The button blends into the background, making it hard for users to notice and use.

### Steps to Reproduce
- Switch the application to light mode.
- Scroll down until the Back to Top button appears.
- Observe that the button is not clearly visible against the light background.

### Expected Behavior

The Back to Top button should remain clearly visible in both light and dark themes while maintaining consistency with the application's design system.

### Proposed Approach
- Update the button's styling for light mode to improve visibility.
- Use theme-aware colors for the background, border, and icon.
- Ensure the button maintains sufficient contrast and remains consistent with the existing UI in both light and dark themes.

### Screenshots

**Dark Mode:** Button is visible
Image

**Light Mode:** Button is not visible properly
Image

### GSSoC 2026

### Code of Conduct

yogita-mehta:
Closes #3051

## What this does
Improves the Back to Top button's visibility in light mode by adding a subtle, neutral (`rgba(0,0,0,0.08)`) 1px outline ring to its box-shadow, layered underneath the existing green glow.

## Why
The button already uses a solid saturated green gradient (`#22C55E` → `#16A34A`) with a white icon — not a theme-dependent color that fails in light mode. However, its shadow was purely green-tinted (`rgba(34,197,94,...)`), which gives the button soft, blurry edges. A green glow around a green button can visually melt into a bright/light background at the edges, which matches the reported symptom.

Adding a thin neutral dark ring gives the button a crisp, defined edge that reads clearly against any background — light or dark — without changing its color scheme or the existing design language.

## Testing
- Verified in browser: scrolled past the 300px show-threshold in both light and dark mode
- Light mode: button now has a clear, crisp edge against the light page background
- Dark mode: outline is subtle enough that the button's appearance is visually unchanged from before
- No other styling, animation, or behavior was touched

### DIFF
--- a/apps/web/app/[locale]/components/BackToTopButton.tsx
+++ b/apps/web/app/[locale]/components/BackToTopButton.tsx
@@ -171,8 +171,8 @@ export default function BackToTopButton() {
                     background: "linear-gradient(180deg, #22C55E 0%, #16A34A 100%)",
                     /* Spec shadow: 0 8px 24px rgba(34,197,94,0.25) */
                     boxShadow: isScrollingBack
-                        ? "0 4px 12px rgba(34,197,94,0.18)"
-                        : "0 8px 24px rgba(34,197,94,0.25), 0 2px 8px rgba(0,0,0,0.10)",
+                        ? "0 4px 12px rgba(34,197,94,0.18), 0 0 0 1px rgba(0,0,0,0.08)"
+                        : "0 8px 24px rgba(34,197,94,0.25), 0 2px 8px rgba(0,0,0,0.10), 0 0 0 1px rgba(0,0,0,0.08)",
                 }}
             >
                 {/* Scroll progress ring */}


yogita-mehta:
Same as #3586 — the failing check (`sahidawa-api#test`) is unrelated to this PR. It's a pre-existing repo-wide CI issue: the workspace packages (`@sahidawa/shared`, `@sahidawa/types`, `@sahidawa/validators`) aren't being built before `sahidawa-api`'s test step runs, causing every test file in `apps/api` to fail with `Cannot find module '@sahidawa/validators'` (or similar) regardless of what the PR touches.

This PR only touches `apps/web/app/[locale]/components/BackToTopButton.tsx` (a 2-line shadow value change) and doesn't touch `apps/api` or the shared packages at all.

PR #3808, opened 2026-07-21
feat(web): cache medication schedules for offline reading

### LINKED ISSUE #3774: [Feature] Implement PWA Offline Read Caching for Medication Schedules
## Problem / Goal
Recently, we added offline queuing support allowing users to *log* medication doses while disconnected from the internet (PR #3768). However, the app currently lacks the ability to *read* schedules offline. 

If a user launches the SahiDawa PWA without internet access, `fetchSchedules()` and `fetchTodaySummary()` fail, rendering the dashboard blank. Patients need to be able to see their medication timings at all times, regardless of network connectivity.

## Expected Behavior & Instructions
We need to implement a robust offline-read caching strategy for medication schedules.
1. When `fetchSchedules()` and `fetchTodaySummary()` are called, they should cache the latest successful response locally (using IndexedDB via our existing `getSyncDB` or a Service Worker strategy).
2. If the network request fails due to offline status, the app should instantly fallback to returning the locally cached schedule data.
3. Show a small visual indicator (e.g., an "Offline Mode" badge) on the Schedule UI when displaying cached data.

## Files to Touch
- `apps/web/lib/scheduleApi.ts`
- `apps/web/app/[locale]/schedule/page.tsx`
- `apps/web/lib/offline/db.ts` (if extending IndexedDB schemas)
- `apps/web/worker/index.js` (if using Service Worker approach)

## Acceptance Criteria
- [ ] Users can view their medication schedule and today's summary while completely offline (airplane mode).
- [ ] The app silently syncs and updates the local cache in the background when the user comes back online.
- [ ] The existing offline `logDose` queue (implemented in PR #3768) must continue to work flawlessly alongside this read cache.

skypank-coder:

## 📋 PR Summary & Link

- **Closes #3774
- **Summary:** `fetchSchedules()` and `fetchTodaySummary()` were plain `fetch()` calls with no fallback, so launching the PWA offline left the schedule dashboard blank. This adds an IndexedDB offline read cache so patients can always see their timings.
  - **`lib/offline/db.ts`:** bumped the sync DB to v3 and added a `readCache` key/value store + best-effort `readCachePut`/`readCacheGet` helpers, kept separate from the existing `pendingScans`/`pendingReports` write queues so PR #3768's offline dose-log queue is untouched.
  - **`lib/scheduleApi.ts`:** on success, cache the response; when `fetch()` itself throws (genuine offline/network failure — a reachable server returning an HTTP error still surfaces as before), fall back to the cached copy. `fetchTodaySummary` returns `fromCache: true` on the fallback path so the UI can flag stale data. Success return shapes are unchanged.
  - **`app/[locale]/schedule/page.tsx`:** shows an "Offline Mode" badge when rendering cached data, and registers a reconnect-retry callback (via the existing `useOfflineStatus` hook) that silently refreshes the schedule and its cache when connectivity returns.
  - Two deliberate choices: (a) cache fallback only on a thrown `fetch()` (airplane mode), not on HTTP errors — avoids serving stale data when the server is reachable, and sidesteps a "failed to fetch" message collision; (b) `fromCache` is only added on the fallback return, keeping success shapes (and the existing `scheduleApi.test.ts`) intact.
  - Scope: `lib/offline/db.ts`, `lib/scheduleApi.ts`, `app/[locale]/schedule/page.tsx`. Not touching `worker/index.js` — this is per-user auth-scoped JSON, better suited to IndexedDB than a Service Worker cache (which we use for static assets like map tiles).

## 🏷️ PR Type

- [x] ✨ `type: feature`

## ✅ Checklist

- [x] My PR has a linked issue (`Closes #3774`)

### DIFF
--- a/apps/web/app/[locale]/profile/page.tsx
+++ b/apps/web/app/[locale]/profile/page.tsx
@@ -6,6 +6,7 @@ import { Link, useRouter } from "@/i18n/routing";
 import { User, ShieldCheck, Bell, ChevronRight, ArrowLeft, LogIn, LogOut } from "lucide-react";
 import ABHABadge from "@/components/ABHABadge";
 import { useSession } from "@/src/components/AuthProvider";
+import { clearReadCache } from "@/lib/offline/db";
 
 const ACCESS_TOKEN_KEY = "sb-access-token";
 
@@ -131,6 +132,9 @@ export default function ProfilePage() {
     };
     const handleSignOut = () => {
         localStorage.removeItem(ACCESS_TOKEN_KEY);
+        // Wipe cached schedules/summary (PHI) so the next person on a shared
+        // device can't read them offline. Fire-and-forget — never block sign-out.
+        void clearReadCache();
         setSession({ status: "guest" });
         router.push("/");
     };
--- a/apps/web/app/[locale]/schedule/page.tsx
+++ b/apps/web/app/[locale]/schedule/page.tsx
@@ -11,8 +11,10 @@ import {
     Pill,
     Plus,
     RefreshCw,
+    WifiOff,
     XCircle,
 } from "lucide-react";
+import { useOfflineStatus } from "@/hooks/useOfflineStatus";
 import { Link } from "@/i18n/routing";
 import { PageHeader } from "../components/PageHeader";
 import Card from "@/components/Card";
@@ -156,11 +158,12 @@ type LoadState =
     | { kind: "loading" }
     | { kind: "authError" }
     | { kind: "networkError"; message: string }
-    | { kind: "ready"; data: TodaySchedule[]; date: string };
+    | { kind: "ready"; data: TodaySchedule[]; date: string; fromCache: boolean };
 
 export default function SchedulePage() {
     const t = useTranslations("schedule");
     const { token, isLoading: authLoading } = useSession();
+    const { registerRetryCallback, unregisterRetryCallback } = useOfflineStatus();
     const [state, setState] = useState<LoadState>({ kind: "loading" });
     const [doseStatus, setDoseStatus] = useState<Record<string, string>>({});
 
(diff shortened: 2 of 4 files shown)


dipexplorer:
Hi @skypank-coder,  there is a **Critical Privacy / Data Leakage** vulnerability in the current implementation that we need to fix before merging.

IndexedDB is shared across the entire origin and persists even after the user logs out. Right now, ur `ReadCacheKey` uses static strings (`"schedules"` and `"todaySummary"`).

If **User A** logs in, their schedule is saved to `"todaySummary"`. If they log out and **User B** uses the same device (very common in rural shared-device setups or family tablets) while the device is **offline**, the fallback logic will trigger and serve User B the medical schedule of User A. In a healthcare app, leaking PHI (Protected Health Information) across user sessions is a massive security violation.

###  Required Fixes
1. **Scope Cache Keys by User ID:** The cache keys must be tied to the currently authenticated user. Instead of a static `"todaySummary"`, append the user's ID or a hash of the token (e.g., `todaySummary_${userId}`). u can decode the `sb-access-token` JWT to extract the `sub` (User ID).
2. **Handle Logout DB Cleanup (Optional but Recommended):** In `app/[locale]/profile/page.tsx`, the `handleSignOut` function only clears the `localStorage`. Plz explore adding a helper to wipe the IndexedDB `readCache` table when the user explicitly signs out.

skypank-coder:
@dipexplorer Thanks for catching this — fixed in the latest commit. Cache entries are now namespaced by the authenticated user's id (decoded from the sb-access-token JWT sub), and the read/write helpers no-op when there's no user, so after logout the offline fallback is disabled entirely rather than serving the previous user's data. I also wired clearReadCache() into handleSignOut to wipe the store on explicit sign-out. Verified with a test reproducing the shared-device scenario (User A caches → logs out → User B offline): it now blocks the read, and I confirmed the same test fails against the pre-fix code (reproducing the leak). Existing schedule/profile tests still pass.

dipexplorer:
quality:exceptional (× 1.5)
Reason: Added edge cases (like offline on a shared device) with deep understanding & proper clearReadCache cleanup implemented on logout.

Issue #3745, opened 2026-07-18
fix(security): Apply missing rate limiters to unprotected endpoints across API routes


dipexplorer:
## Description

Recent CodeQL Security scans have flagged multiple API endpoints across the application as lacking rate limit protection (Alerts #175, #174, #173, #172, #153, etc.). 

While previous PRs (like those closing #2617, #2618, #2619, #2620, #2621) introduced rate limiters (`limiter`, `reportLimiter`, `authTargetLimiter`, etc.) and applied them to several routes, a significant number of endpoints were unintentionally missed during that implementation.

**Missing endpoints identified so far:**
- `apps/api/src/routes/reports.ts`: `GET /mine`, `GET /`, `PATCH /:id/status`
- `apps/api/src/routes/notifications.ts`: `POST /subscriptions`, `DELETE /subscriptions`, `POST /recalls/mock/trigger`, `GET /status`, `DELETE /phone`, `POST /broadcast`
- Missing endpoints also exist in `wishlist.ts`, `pharmacies.ts`, `analytics.ts` as flagged by CodeQL.

## Goal
Go through the affected route files and ensure that *every* route has an appropriate rate limiter applied to it. 

## Files to Touch
- `apps/api/src/routes/reports.ts`
- `apps/api/src/routes/notifications.ts`
- `apps/api/src/routes/wishlist.ts`
- `apps/api/src/routes/pharmacies.ts`
- `apps/api/src/routes/analytics.ts`
- `apps/api/src/app.ts` (for global routes if any)

## Acceptance Criteria
- [ ] Inspect each endpoint in the listed files.
- [ ] Import the appropriate limiter from `apps/api/src/middleware/rateLimit.ts` (e.g., `limiter` for general endpoints, `reportLimiter` for reports, etc.).
- [ ] Apply the limiter as middleware to every missing endpoint.
- [ ] Ensure that adding the limiter does not break existing authentication middlewares.

### ⚠️ Contributor Instructions (Please Read Carefully)

- **Do NOT spam "/assign" or "Please assign me".**
- To claim this task, you MUST reply with a brief **proposed implementation plan/approach**. What files will you touch? How will you solve it?
- Once your approach is reviewed and approved by a maintainer, you will be officially assigned.
- Any PR opened without prior assignment and approach approval will be closed.

Avinash-sdbegin:
Hi @dipexplorer,

I'd like to work on this issue.

### Proposed implementation plan

[...]

PR #3811, opened 2026-07-21
fix(api): stop notification endpoints leaking OTP and user_id in responses

### LINKED ISSUE #3810: [BUG] Broken access control in GET /api/notifications/status exposes subscriber PII and OTP to unauthenticated callers

### Bug Description

`GET /api/notifications/status` (mounted at both `/api/notifications` and `/api/v1/notifications`) runs behind `optionalAuth`, so it answers unauthenticated callers too. On the guest path it looks a subscriber up by a `phone` value read straight from the query string and hands back the whole row via `select("*")`:

```ts
// apps/api/src/routes/notifications.ts:182-242
let query = supabase.from("notification_subscribers").select("*");
if (req.user) {
  query = query.eq("user_id", req.user.id);
} else if (phone) {
  query = query.eq("phone", phone);   // guest path — no auth
}
...
res.json({ registered: true, subscriber }); // full row returned
```

Two things turn this into a live bug rather than a theoretical one:

1. The API's Supabase client is built with `SUPABASE_SERVICE_ROLE_KEY` (`apps/api/src/db/client.ts`), which bypasses Row Level Security. The `notification_subscribers` RLS policy that limits rows to `user_id = auth.uid()` does nothing here. The query scoping in this route is the only access control, and the guest path has none.
2. `notification_subscribers` stores `verification_otp`, `otp_expires_at`, `user_id`, `district`, `phone`, and `channels` (see `supabase/migrations/20260611000002_add_notification_subscribers.sql` and `20260625120000_add_notification_verification.sql`).

**Impact**

With no login at all, anyone can look up any phone number and read that subscriber's full record back:

```
GET /api/notifications/status?phone=%2B919876543210
→ 200 { "registered": true, "subscriber": { "user_id": "...", "district": "...", "verification_otp": "...", ... } }
```

- For any subscriber this exposes the Supabase `user_id` (account UUID), district, phone, and channels — an unauthenticated phone → account lookup over health users.
- When a pending verification is in flight, the same response also carries the plaintext `verification_otp`. That undercuts the phone-ownership check: someone can register a number they don't own (`POST /register`), read the OTP back off `/status`, then finish `/verify-otp`.

### Expected Behavior

`/status` should never return sensitive columns to an unauthenticated caller.

- The guest-by-phone path should reply with existence only — `{ "registered": true | false }` — and never the row itself.
- Swap `select("*")` for an explicit column list that leaves out `verification_otp` and `otp_expires_at`.
- Worth considering: require auth on `/status` and drop the guest-by-phone lookup entirely.

### Related

Separate from #3745, which adds rate limiters to `/status` and other routes. This one is about what the endpoint returns, not how often it can be called.

### GSSoC 2026

### Code of Conduct

- [x] I have read and agree to SahiDawa's Code of Conduct.

shashank03-dev:

## 📋 PR Summary & Link

- **Closes #3810**
- **Summary:**

`GET /api/notifications/status` runs under `optionalAuth`, so it also answers callers who aren't logged in. On the guest path it looked a subscriber up by a `phone` query param and returned the entire row via `select("*")`, which includes `verification_otp`, `otp_expires_at` and the Supabase `user_id`. The API talks to Supabase with the service-role key, so RLS is bypassed and the query scoping in this route is the only access control there is. The guest path had none, so anyone could read a subscriber's OTP and account UUID just by knowing their phone number.

`POST /register` and `PATCH /phone` returned the same full row. Worse, `register` handed the freshly generated OTP straight back in its own response, which defeats the point of the phone-ownership check: register a number you don't own, read the OTP off the response, then verify.

The fix is a small `toPublicSubscriber()` projection that only exposes the fields a client actually uses (`phone`, `channels`, `language`, `district`, `is_active`), applied everywhere a subscriber is returned. `/status` now selects only those columns too, so the OTP isn't even pulled out of the database on that path.

There's no API contract change here. The web client's `SubscriberData` type already reads only those five fields, so nothing on the frontend has to move, and there's no schema or migration.

This is scoped to the data disclosure only. It's separate from #3745, which is about rate limiting the same routes.

Backend-only change, no UI. The added tests assert the sensitive fields never show up in any response: authenticated `/status`, unauthenticated guest `/status`, `/register`, and `PATCH /phone`.

```
PASS apps/api/tests/notifications.test.ts
  notifications routes
    ✓ fetches subscription status successfully
    ✓ never leaks OTP or user_id from /status for an authenticated user
    ✓ never leaks OTP or user_id from /status for an unauthenticated guest lookup
    ✓ registers a subscriber successfully
    ✓ updates subscriber details successfully
    ...
Tests:       19 passed, 19 total
```

## 🏷️ PR Type

- [x] 🐛 `type: bug`
- [ ] ✨ `type: feature`
- [ ] 📖 `type: docs`
- [ ] 🧪 `type: testing`
- [x] 🔒 `type: security`
- [ ] ⚡ `type: performance`
- [ ] 🎨 `type: design`
- [ ] ♻️ `type: refactor`
- [ ] 🛠️ `type: devops`
- [ ] ♿ `type: accessibility`

## ✅ Checklist

- [x] My PR has a linked issue (`Closes #3810`)

### DIFF
--- a/apps/api/src/routes/notifications.ts
+++ b/apps/api/src/routes/notifications.ts
@@ -168,6 +168,27 @@ interface InMemorySubscriber {
 // (Live view of dbConfig.isSupabaseOffline is read directly inside request handlers)
 const memorySubscribers = new Map<string, InMemorySubscriber>();
 
+// The only subscriber fields any client is allowed to receive. Everything else on
+// a row is either a secret (verification_otp, otp_expires_at) or account-linkage
+// PII (user_id) and must never be serialized into a response. Returning the raw
+// row from /status let an unauthenticated caller read another subscriber's OTP and
+// Supabase user_id just by knowing their phone number, so every subscriber we hand
+// back — from the DB or the in-memory fallback — goes through this projection.
+type PublicSubscriber = Pick<
+    InMemorySubscriber,
+    "phone" | "channels" | "language" | "district" | "is_active"
+>;
+
+function toPublicSubscriber(sub: PublicSubscriber): PublicSubscriber {
+    return {
+        phone: sub.phone,
+        channels: sub.channels,
+        language: sub.language,
+        district: sub.district,
+        is_active: sub.is_active,
+    };
+}
+
 router.get("/status", optionalAuth, async (req: AuthenticatedRequest, res) => {
     try {
         let phone: string | undefined = undefined;
@@ -179,7 +200,9 @@ router.get("/status", optionalAuth, async (req: AuthenticatedRequest, res) => {
             }
             phone = formatted;
         }
-        let query = supabase.from("notification_subscribers").select("*");
+        let query = supabase
+            .from("notification_subscribers")
+            .select("phone, channels, language, district, is_active");
 
         if (req.user) {
             query = query.eq("user_id", req.user.id);
@@ -239,7 +262,7 @@ router.get("/status", optionalAuth, async (req: AuthenticatedRequest, res) => {
             return;
         }
 
-        res.json({ registered: true, subscriber });
+        res.json({ registered: true, subscriber: toPublicSubscriber(subscriber) });
     } catch (err) {
         logger.error({ message: "Error in /status endpoint", error: err });
         res.status(500).json({ error: "Internal server error" });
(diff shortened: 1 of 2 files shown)


dipexplorer:
quality:exceptional (× 1.5 multiplier) - The contributor wrote excellent security-focused test cases to prove that the sensitive fields are completely scrubbed from all HTTP responses.

PR #3815, opened 2026-07-21
Fix: Empty scan results cached for 1 hour, hiding newly added medicines from users

### LINKED ISSUE #3793: [BUG] Empty scan results cached for 1 hour, hiding newly added medicines from users

### Bug Description

In `apps/api/src/routes/scan.ts:428`, when the fuzzy brand matching query returns zero results, an empty array `[]` is cached in Redis with a 1-hour TTL (`{ EX: 3600 }`). If new medicines are added to the database during that hour (e.g., via the ETL pipeline or admin data import), users scanning those barcodes will continue to see no results until the cache expires or is manually invalidated.

This creates a stale data problem where the scanning feature returns false negatives for up to an hour after legitimate data additions.

### Expected Behavior

Empty results should not be cached at all, or should have a significantly shorter TTL (e.g., 60 seconds) to ensure newly added medicines become discoverable quickly.

### Screenshots

_No response_

### GSSoC 2026

### Code of Conduct

Kirtan-pc:

> [!WARNING]
> PRs with unrelated files will not be reviewed and may be closed.

## 📋 PR Summary & Link

- **Closes #3793 **
- **Summary:**
1. File Persistence (apps/api/src/services/memorySubscriberStore.ts):
- Serializes the store to data/memory-subscribers.json on every mutation (2s debounce)
- Loads from disk on startup — data survives server restarts
- Periodic safety save every 30s
2. Automatic Supabase Reconciliation:
- Runs every 30s (matching the health probe interval)
- When Supabase is online and items exist in the local store, syncs each subscriber via find-then-update-or-insert
- Successfully reconciled subscribers are removed from the local store
- Partial reconciliation is safe — failures are logged per-subscriber
3. Gitignore: Added memory-subscribers.json to data/.gitignore since it is runtime-generated state.

## 🏷️ PR Type

- [x] 🐛 `type: bug`
- [ ] ✨ `type: feature`
- [ ] 📖 `type: docs`
- [ ] 🧪 `type: testing`
- [ ] 🔒 `type: security`
- [ ] ⚡ `type: performance`
- [ ] 🎨 `type: design`
- [ ] ♻️ `type: refactor`
- [ ] 🛠️ `type: devops`
- [ ] ♿ `type: accessibility`

## ✅ Checklist

- [x] My PR has a linked issue (`Closes #3793`)


dipexplorer:
@Kirtan-pc, **Build Failure (TypeScript):** In `apps/api/src/routes/notifications.ts`, it looks like `type PublicSubscriber` and `function toPublicSubscriber` were accidentally pasted twice (around line 159 and again at 180). This is causing `tsc` to fail with a `Duplicate function implementation` error. Plz remove the duplicated block.

Kirtan-pc:
Hey @dipexplorer,
I have fixed it,
Could you please review it again?

dipexplorer:
Hey @Kirtan-pc! Thanks for the update

The logic looks solid conceptually, but there are 2 critical Node.js anti-patterns that need to be fixed before production:

1. Event Loop Blocking (Synchronous I/O): u're using fs.writeFileSync inside save(). Since this runs in an Express API, writing a JSON array to disk synchronously will block entire Node.js event loop, freezing all other incoming API requests until write completes. Plz refactor this to use asynchronous fs.promises.writeFile.
2. Reconciliation Race Condition: In startReconciliation(), u're passing an async function to setInterval(..., 30_000). If Supabase is slow and reconciliation takes longer than 30 seconds, next interval will fire & overlap with the currently running one, causing race conditions, duplicated database calls. Plz add an isReconciling boolean flag to skip the next interval if a sync is already in progress.

 push these fixes & I'll approve and merge!

Kirtan-pc:
Hey @dipexplorer,
I have implemented the changes:
1. Async file I/O - save() now uses fsp.writeFile (from fs/promises) instead of fs.writeFileSync. All callers (debouncedSave, startPeriodicSave, reconcileWithSupabase, saveImmediate) properly handle the returned promise with .catch() for fire-and-forget paths.
2. isReconciling guard - Added a private isReconciling = false flag to startReconciliation(). If a reconciliation cycle is already in progress when the next 30s interval fires, it returns early. The flag is set true before await this.reconcileWithSupabase() and reset in a finally block.
Could you please review it?

PR #3991, opened 2026-08-01
fix: replace console.log with structured logger in tracing.ts

sanrishi:
Replaces console.log with logger.info/error for tracing shutdown messages.


dipexplorer:
this change has already been implemented and merged as part of PR #3955.

PR #4044, opened 2026-08-03
fix(districtAlertSync): retry and throw on stale alert deactivation failure

### LINKED ISSUE #4041: Bug: Stale district alerts remain active when bulk deactivation fails in districtAlertSync
### Description
In `districtAlertSync.ts`, when the bulk deactivation of stale district alerts fails (lines 131-142), the error is only logged but the stale alerts remain active in the database. Users continue to see outdated/stale district alerts even after they should have been deactivated. The function should retry or at least ensure stale alerts are marked inactive to prevent serving stale data.

### Affected File
`apps/api/src/cron/districtAlertSync.ts:131-142`

### Root Cause
```typescript
if (staleIds.length > 0) {
    const { error: updateError } = await supabase
        .from("district_alerts")
        .update({ is_active: false, updated_at: timestamp })
        .in("id", staleIds);

    if (updateError) {
        logger.error("District alert sync: bulk deactivate failed", {
            error: updateError.message,
        });
        errors += 1;
        // BUG: stale alerts remain is_active: true in the database!
    }
}
```

### Reproduction Steps
1. Have stale district alerts in the database (alerts with no remaining verified reports)
2. Cause the bulk deactivation to fail (e.g., temporary DB outage, network issue)
3. Observe that stale alerts remain `is_active: true` and are served to users

### Expected Behavior
When bulk deactivation fails, the function should retry or throw an error to prevent stale alerts from being served. At minimum, the error should be surfaced so the stale alerts are not silently left active.

### Fix
Add retry logic or throw an error when deactivation fails, so stale alerts are not silently left active. Alternatively, log the specific stale alert IDs so they can be manually cleaned up.

### Labels
bug, gssoc2026, backend

vipul674:
Fixes #4041

- Retry bulk deactivation once when it fails
- Throw error if retry also fails to prevent silent data corruption
- Ensures stale alerts are not left active in the database

### DIFF
--- a/apps/api/src/cron/districtAlertSync.ts
+++ b/apps/api/src/cron/districtAlertSync.ts
@@ -134,10 +134,21 @@ export async function syncDistrictAlertTallies(): Promise<void> {
                 .in("id", staleIds);
 
             if (updateError) {
-                logger.error("District alert sync: bulk deactivate failed", {
+                logger.error("District alert sync: bulk deactivate failed, retrying", {
                     error: updateError.message,
                 });
-                errors += 1;
+                const { error: retryError } = await supabase
+                    .from("district_alerts")
+                    .update({ is_active: false, updated_at: timestamp })
+                    .in("id", staleIds);
+                if (retryError) {
+                    logger.error("District alert sync: bulk deactivate retry failed", {
+                        error: retryError.message,
+                    });
+                    throw new Error(
+                        `Failed to deactivate stale district alerts: ${retryError.message}`
+                    );
+                }
             }
         }
 


vipul674:
@dipexplorer

## Fix Report

**Issue:** #4041 — Stale district alerts remain active when bulk deactivation fails in districtAlertSync

**Root Cause:** When bulk deactivation of stale district alerts fails, the error is only logged but stale alerts remain  in the database. Users continue seeing outdated alerts.

**Fix:** Added retry logic — on first failure, retry once. If retry also fails, throw an error instead of silently leaving stale alerts active.

**Files changed:** 1 ()

**Impact:** Ensures stale alerts are either deactivated or surfaced as an error, preventing silent data corruption.

PR #4045, opened 2026-08-03
fix(interactions): default isFallback to false instead of true

### LINKED ISSUE #4042: Bug: Drug interaction endpoint marks all results as unverified when dbConfig is undefined
### Description
In `interactions.ts`, the `isFallback` variable defaults to `true` when `dbConfig` is undefined or `dbConfig.isSupabaseOffline` is undefined due to the `?? true` operator. This causes all drug interaction results to be marked as `verified: false` even when the database is online and serving live data, misleading users about the reliability of interaction data.

### Affected File
`apps/api/src/routes/interactions.ts:388,640`

### Root Cause
```typescript
const isFallback = dbConfig?.isSupabaseOffline ?? true;
// ...
verified: !isFallback,  // false when dbConfig is undefined!
```

The `?? true` default is overly pessimistic. When `dbConfig` is undefined (e.g., during module initialization or in certain code paths), the system incorrectly assumes fallback mode and reports all interactions as unverified.

### Reproduction Steps
1. Start the API server with a properly configured Supabase connection
2. Query the interactions endpoint (`GET /api/v1/interactions?ids=...`)
3. Observe that `verified` is `false` in the response even though the database is online and serving live data

### Expected Behavior
When `dbConfig` is undefined or `isSupabaseOffline` is undefined, the system should default to assuming the database is available (i.e., `isFallback = false`), since the presence of `dbConfig` indicates the application is running with a configured database.

### Fix
Change the default from `true` to `false`:
```typescript
const isFallback = dbConfig?.isSupabaseOffline ?? false;
```

### Labels
bug, gssoc2026, backend

vipul674:
Fixes #4042

- Change `dbConfig?.isSupabaseOffline ?? true` to `?? false`
- Prevents all interactions from being marked as unverified when dbConfig is undefined

### DIFF
--- a/apps/api/src/routes/interactions.ts
+++ b/apps/api/src/routes/interactions.ts
@@ -385,7 +385,7 @@ router.get(
             const interactionByPair = indexInteractions(
                 await loadInteractionsForGenerics(selectedGenerics)
             );
-            const isFallback = dbConfig?.isSupabaseOffline ?? true;
+            const isFallback = dbConfig?.isSupabaseOffline ?? false;
             const interactions = [];
 
             for (let i = 0; i < medicines.length; i++) {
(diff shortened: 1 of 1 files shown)


vipul674:
@dipexplorer

## Fix Report

**Issue:** #4042 — Drug interaction endpoint marks all results as unverified when dbConfig is undefined

**Root Cause:**  defaults to  when dbConfig is undefined, causing all interactions to be marked  even when the database is online.

**Fix:** Changed default from  to : 

**Files changed:** 1 ()

**Impact:** Interactions are correctly marked as verified when the database is online and available.

PR #4105, opened 2026-08-04
fix: sahidawa pharmacies promise catch

sanrishi:
Adds proper error handling for sahidawa pharmacies promise catch.


dipexplorer:
@sanrishi plz raise the issue against the PR u r working.

PR #4144, opened 2026-08-05
feat(api): add page/limit pagination to GET /api/wishlist

### LINKED ISSUE #4048: [FEATURE] Add pagination to GET /api/wishlist
### Which area does this feature belong to?

Backend — API / Express (apps/api)

### Estimated Difficulty

🟡 Intermediate — Some experience needed

### What problem does this solve?

GET /api/wishlist in apps/api/src/routes/wishlist.ts currently fetches a user's entire wishlist in one query with no page/limit parameters — it just orders by created_at descending and returns everything. Every other list-style endpoint in the API that can grow large (e.g. GET /api/v1/alerts) already implements page/limit pagination with sane defaults and a max page size, so this endpoint is inconsistent with the rest of the codebase and will slow down or return unbounded payloads for users with large wishlists.

### Proposed Solution

Add page (default 1) and limit (default 20, capped at 100) query params, apply .range() on the Supabase query the same way alerts.ts does, and return { items, count, page, limit, totalPages } instead of the current flat { success, count, items } shape — noting in the PR that this is a response-shape change requiring a matching frontend update. The endpoint should honor page/limit, default sensibly when they're absent or invalid, and the total count should still reflect the full wishlist size, not just the current page.

### Acceptance Criteria

_No response_

### GSSoC 2026

- [x] I am a GSSoC 2026 participant and would like to implement this feature.
- [ ] This feature uses **Cloudinary** (eligible for GSSoC Cloudinary bounty 🏆).

### Code of Conduct

OmanshiRaj:
**Closes #4048 **
- **Summary:** `GET /api/wishlist` previously fetched a user's entire wishlist in
  one unbounded query. This adds `page` (default 1) and `limit` (default 20,
  capped at 100) query params, applies `.range()` on the Supabase query the
  same way `alerts.ts` does, and returns `{ items, count, page, limit,
  totalPages }` instead of the old flat `{ success, count, items }` shape.
  `count` reflects the full wishlist size across all pages, not just the
  current page.

### DIFF
--- a/apps/api/tests/wishlist.test.ts
+++ b/apps/api/tests/wishlist.test.ts
@@ -1,4 +1,3 @@
-// @ts-nocheck
 import express from "express";
 import request, { Response as SuperTestResponse } from "supertest";
 import wishlistRouter, { mergeGuestWishlist } from "../src/routes/wishlist";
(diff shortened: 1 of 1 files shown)


dipexplorer:
Tests cannot pass because actual GET `/api/v1/wishlist` route does not implement page and limit query parameter support yet. it must
1. Accept page & limit query params.
2. Return the response shape { items, page, limit, count, totalPages }.

Plz add the implementation, corresponding route/controller file alongside these tests.

PR #4204, opened 2026-08-07
fix: sahidawa settings promise catch

sanrishi:
Adds proper error handling for sahidawa settings promise catch.


dipexplorer:
@sanrishi, adds a second .catch() block immediately after an existing .catch() block in the promise chain. Since the first catch block handles the error and does not re-throw it, this new second catch block becomes unreachable dead code. Plz review the promise chain logic and remove this redundant block.

PR #4244, opened 2026-08-09
fix(alert-broadcaster): prevent duplicate broadcasts with lock renewal and in-process guard

### LINKED ISSUE #4242: [BUG] Alert-broadcaster Redis lock expires mid-run → overlapping broadcasts → duplicate patient alerts

Other / Unknown

### Bug Description

broadcast lock is `LOCK\\\_TTL\\\_MS = 25\\\_000` but the scheduler interval is `CHECK\\\_INTERVAL\\\_MS = 30\\\_000`:
```ts

const LOCK\\\_TTL\\\_MS = 25\\\_000;  // <25s TTL, no renewal

...

setInterval(() => { checkAndBroadcastAll().catch(...) }, 30\\\_000);

```
A single run pages every subscriber via 50-item `Promise.allSettled` chunks of SMS/WhatsApp; any slow provider call pushes the run past 25s. The lock expires (no renewal) and the next tick — or a second pod — acquires the lock and runs **concurrently** over the same `broadcasted=false` rows. There is no in-process `isRunning` guard, so even a single instance can overlap.

**Impact:** Duplicate counterfeit/recall/expiry SMS + WhatsApp to every subscriber, at real per-message cost, on every slow broadcast window.

### Expected Behavior

Add an in-process run guard + refresh/replenish the lock TTL during the run (or use Redlock with renewal) so acquisition is exclusive.

### Screenshots

_No response_

### GSSoC 2026

### Code of Conduct

Kirtan-pc:

> [!WARNING]
> PRs with unrelated files will not be reviewed and may be closed.

## 📋 PR Summary & Link

- **Closes #4242**
- **Summary:**
In apps/api/src/cron/alert-broadcaster.ts:
1. Lock TTL renewal — new renewLock() runs a compare-and-PEXPIRE Lua script on a setInterval heartbeat (LOCK_RENEW_INTERVAL_MS = LOCK_TTL_MS / 2). The TTL is extended only while this process still owns the lock, so a slow run can't hand it to another tick/pod. Old releaseLock() (del only-if-owner) is unchanged for the graceful-cleanup path.
2. In-process run guard — new module-level isBroadcasting flag set true after acquireLock() and reset in finally. checkAndBroadcastAll() now returns early if a run is already in flight in this process, before even attempting the distributed lock.
This makes acquisition exclusive both across pods (renewed Redis lock) and within a pod (guard), eliminating duplicate sends even on slow broadcast windows. The renewal timer is unref()'d so it never keeps the process alive.
Tests
Extended alert-broadcaster.test.ts (node:test / mock.module style):
- skips a second checkAndBroadcastAll() while a run is in progress (guard)
- renews the lock TTL multiple times during a run that outlives the original TTL
Existing distributed-lock tests updated accordingly and passing behavior preserved. Jest collection isn't wired to src/cron/*.test.ts in jest.config.js, so these run standalone via the Node test runner.

## 🏷️ PR Type

- [x] 🐛 `type: bug`
- [ ] ✨ `type: feature`
- [ ] 📖 `type: docs`
- [ ] 🧪 `type: testing`
- [ ] 🔒 `type: security`
- [ ] ⚡ `type: performance`
- [ ] 🎨 `type: design`
- [ ] ♻️ `type: refactor`
- [ ] 🛠️ `type: devops`
- [ ] ♿ `type: accessibility`

## ✅ Checklist

- [x] My PR has a linked issue (`Closes #4242 `)

### DIFF
--- a/apps/api/src/cron/alert-broadcaster.test.ts
+++ b/apps/api/src/cron/alert-broadcaster.test.ts
@@ -1,4 +1,7 @@
 // @ts-nocheck
+// Force the test branch of module constants (LOCK_RENEW_INTERVAL_MS, etc.)
+process.env.NODE_ENV = "test";
+
 // Fix: Mock WebSocket for Node.js versions < 22 to prevent Supabase Realtime client crash during test imports
 if (typeof globalThis.WebSocket === "undefined") {
     globalThis.WebSocket = class {} as any;
@@ -25,6 +28,10 @@ const mockRedisClient = {
             const [key] = keys;
             const [expected] = args;
             if (lockStore.get(key) === expected) {
+                if (args.length > 1) {
+                    // renew script: PEXPIRE — keep the lock alive
+                    return 1;
+                }
                 lockStore.delete(key);
                 return 1;
             }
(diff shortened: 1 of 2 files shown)


dipexplorer:
"quality: exceptional: for implementation of a distributed lock heartbeat via atomic Lua scripts

PR #4271, opened 2026-08-10
fix(api): stem-match emergency keywords so inflected phrasing isn't missed

### LINKED ISSUE #4270: [Bug] Emergency keyword detection misses common word variants (e.g. "suicidal" doesn't match "suicide")

### Bug Description

Which part of SahiDawa is broken?
Backend — Triage/RAG service (apps/api/src/services/medicineRag.service.ts)

Bug Description
assessUrgency() decides whether a symptom query gets flagged as a medical emergency (triggering the urgent-care disclaimer in /api/triage/recommend). It works via plain substring matching against a fixed keyword list:

const EMERGENCY_KEYWORDS = [
    "chest pain", "heart attack", "stroke", "unconscious", "not breathing",
    "difficulty breathing", "shortness of breath", "severe bleeding",
    "suicide", "seizure", "paralysis", "fainting", "high fever",
];

export function assessUrgency(text: string): UrgencyAssessment {
    const normalized = text.toLowerCase();
    const matched = EMERGENCY_KEYWORDS.filter((keyword) => normalized.includes(keyword));
    return { emergency: matched.length > 0, matched };
}

Because this is exact substring matching, it silently misses extremely common natural phrasing of the same concept — most seriously:

- "I feel suicidal" does NOT match "suicide" — "suicidal" and "suicide" diverge at the 7th character ("suicid" + "a" vs "suicid" + "e"), so this is a false negative on the single most safety-critical keyword in the list.
- "can't breathe" doesn't match any of "not breathing", "difficulty breathing", or "shortness of breath".
- "fainted" doesn't match "fainting".

The existing test suite (apps/api/tests/triage.test.ts) only checks exact-phrase inputs like "I have severe chest pain", so this gap isn't caught anywhere.

### Expected Behavior

Expected Behavior
assessUrgency should catch common real-world phrasings of the same emergency, not just the exact keyword string — at minimum via stemmed/prefix matching (e.g. match on "suicid" instead of "suicide") or a small synonym/variant map per keyword, so a message like "I feel suicidal" correctly sets emergency: true and returns the urgent-care disclaimer instead of a normal medicine suggestion.

Files to touch
- apps/api/src/services/medicineRag.service.ts
- apps/api/tests/triage.test.ts (add coverage for variant forms)

### Screenshots

_No response_

### GSSoC 2026

### Code of Conduct

skypank-coder:

## 📋 PR Summary & Link

- **Closes #** <issue-number-if-any>
- **Summary:** `assessUrgency` in `apps/api/src/services/medicineRag.service.ts` used exact substring matching against `EMERGENCY_KEYWORDS`, producing dangerous false negatives on common phrasing — most critically **"I feel suicidal" did not match "suicide"** (they diverge at the 7th character), and "can't breathe" / "fainted" matched nothing either.
  - Replaced the `includes()` check with **word-boundary regex matching**. `EMERGENCY_KEYWORDS` is now a list of `{ keyword, pattern }`, pairing each canonical label with a stem/prefix pattern:
    - `\bsuicid` → suicidal, suicide(s), suicided
    - `\bfaint` → faint, fainted, fainting, faints
    - `\bbreath` → not breathing, difficulty breathing, shortness of breath, **and** can't / cannot / cant breathe (all start with "breath")
    - `\bseizur` → seizure(s); `\bparaly` → paralysis, paralyzed, paralyzing
  - The leading `\b` stops a stem from matching inside an unrelated word — so `"keystroke"` no longer trips `"stroke"` (which the old `includes()` **would** have).
  - `matched` still returns the **readable canonical keyword** (not the stem), so the `UrgencyAssessment` shape (`{ emergency, matched }`) is unchanged and `triage.ts`'s `urgentKeywords` consumer (validated as `z.array(z.string())`) is unaffected. No API contract or route changes.
  - Design note: for emergency triage, false positives (over-flagging) are far safer than the false negatives being fixed here, so the broad `breath` stem is an intentional, safety-biased trade-off.
  - Scope: `apps/api/src/services/medicineRag.service.ts` and its test file `apps/api/tests/triage.test.ts` only.

Added tests for every fixed variant plus a regression guard, and re-ran the triage suite:

New assertions:
- `emergency: true` for "I feel suicidal", "can't breathe", "she fainted", "having a seizure right now", "left arm is paralyzed".
- `matched` returns canonical keywords ("suicide", "difficulty breathing", "paralysis"), not stems.
- `"I pressed the wrong keystroke"` → `emergency: false` (word-boundary guard).
- Existing cases unchanged: "I have severe chest pain" → true, "mild headache and runny nose" → false.

## 🏷️ PR Type

- [x] 🐛 `type: bug`

## ✅ Checklist

- [x] My PR has a linked issue Closes #4270

### DIFF
--- a/apps/api/src/services/medicineRag.service.ts
+++ b/apps/api/src/services/medicineRag.service.ts
@@ -35,20 +35,30 @@ export const MEDICINE_RAG_DISCLAIMER =
  * self-medicate. Kept intentionally small and language-light; the voice flow
  * already runs richer multilingual emergency detection on the client.
  */
-const EMERGENCY_KEYWORDS = [
-    "chest pain",
-    "heart attack",
-    "stroke",
-    "unconscious",
-    "not breathing",
-    "difficulty breathing",
-    "shortness of breath",
-    "severe bleeding",
-    "suicide",
-    "seizure",
-    "paralysis",
-    "fainting",
-    "high fever",
+// Each keyword pairs a canonical label (returned in `matched`, so consumers keep
+// seeing readable words rather than stems) with a word-boundary regex. Stems are
+// used where common inflected forms would otherwise be missed — e.g. "suicidal"
+// did not match a plain "suicide" substring, and "can't breathe" did not match
+// any breathing phrase. The leading \b avoids matching a stem inside an unrelated
+// word (e.g. "keystroke" no longer trips "stroke").
+const EMERGENCY_KEYWORDS: { keyword: string; pattern: RegExp }[] = [
+    { keyword: "chest pain", pattern: /\bchest pain/ },
+    { keyword: "heart attack", pattern: /\bheart attack/ },
+    { keyword: "stroke", pattern: /\bstroke/ },
+    { keyword: "unconscious", pattern: /\bunconscious/ },
+    // "breath" stem covers "not breathing", "difficulty breathing", "shortness of
+    // breath", and "can't"/"cannot"/"cant breathe" (all start with "breath").
+    { keyword: "difficulty breathing", pattern: /\bbreath/ },
+    { keyword: "severe bleeding", pattern: /\bsevere bleeding/ },
+    // "suicid" stem → suicidal, suicide(s), suicided.
+    { keyword: "suicide", pattern: /\bsuicid/ },
+    // "seizur" stem → seizure(s).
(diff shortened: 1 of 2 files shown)


dipexplorer:
quality: exceptional: for implementing robust stem-matching using regex with word boundaries

PR #4292, opened 2026-08-11
fix: prevent Overpass cache stampede

### LINKED ISSUE #4289: [Performance] Prevent Overpass cache stampede and unnecessary multi-mirror fan-out
### 🔍 Have You Searched Existing Issues?

- [x] I have searched the existing issues to avoid duplicates

## Problem

The `/api/overpass` proxy uses Redis caching to reduce repeated Overpass API requests. However, when multiple users send the same query concurrently while the response is not yet cached, each request can independently encounter a cache miss and start its own upstream fetch.

Because the proxy may query multiple Overpass mirrors for a single cache miss, a burst of identical requests can generate unnecessary upstream traffic.

For example:

100 concurrent identical requests
→ 100 Redis cache misses
→ multiple upstream Overpass requests per client
→ potentially hundreds of unnecessary requests

This creates a cache stampede/thundering-herd problem and can cause increased latency, unnecessary load on external Overpass infrastructure, and higher risk of rate limiting during traffic spikes.

The expected behavior is that concurrent identical cache misses should be coalesced so that one request performs the upstream fetch, stores the result in Redis, and the other requests reuse the resulting cached response.

The existing mirror fallback and caching behavior should remain functional.

### 🧪 Environment Details

- Application: SahiDawa
- Area: Overpass API proxy / map data fetching
- Component: `/api/overpass`
- Runtime: Next.js / Node.js
- Caching: Redis
- External dependency: Overpass API mirrors
- Reproduction scenario: Multiple concurrent requests for the same uncached Overpass query
- Impact: Increased upstream requests, unnecessary network usage, latency spikes, and potential rate-limit pressure
- Expected: Concurrent identical requests should share/coalesce a single upstream fetch

### 🔁 Steps to Reproduce

1. Open the SahiDawa application and navigate to a page that uses the Overpass map-data proxy.

2. Identify an Overpass query that is not currently present in the Redis cache.

3. Send the same Overpass query concurrently from multiple clients/requests.

4. Observe that each request can independently encounter a Redis cache miss.

5. Monitor the outgoing network requests to the configured Overpass mirrors.

6. Notice that multiple identical upstream requests are initiated instead of sharing a single in-flight request.

7. Repeat the test with a larger concurrent burst (for example, 20–100 identical requests) to observe the increased upstream request fan-out.

8. Compare this behavior with the expected result where one request performs the upstream fetch, stores the response in Redis, and concurrent requests reuse that result.

### 📋 Logs / Screenshots (Optional)

No specific error is required to reproduce this issue.

The most useful evidence would be:

- Network/request logs showing multiple identical Overpass requests for the same query.
- Redis cache-miss logs for concurrent identical requests.
- Upstream request counts during a concurrent burst.
- Timing information showing multiple requests waiting on separate upstream fetches.

A load/concurrency test demonstrating that 20–100 identical requests can trigger multiple upstream fetches would provide strong evidence of the cache-stampede behavior.

### 🙌 Contributor Checklist

- [x] I agree to follow this project's Code of Conduct
- [x] I want to work on this issue
- [x] I am a GSSOC'26 contributor

Divyanshu3994:

- [x] I verified that this PR ONLY touches the required file.

> [!WARNING]
> PR contains only the changes required for this issue. No unrelated files have been modified.

## 📋 PR Summary & Link

- **Closes #4289 
- **Summary:**
  - Prevented cache stampede in the Overpass API route.
  - Added query-specific Redis distributed locking using atomic `SET NX` with TTL.
  - Concurrent identical requests now coalesce around a single upstream fetch instead of independently querying Overpass mirrors.
  - Added a cache double-check after lock acquisition to prevent unnecessary upstream requests.
  - Added bounded cache polling for requests waiting on an existing lock.
  - Implemented token-safe atomic lock release using Redis Lua `EVAL`.
  - Preserved the existing Overpass mirror fetching and caching behavior.
  - Added Redis failure handling so the API can fall back to the existing upstream behavior.

> [!IMPORTANT]
> **No Pull Request will be merged without proof of testing!**

### Testing / Verification

- The implementation was reviewed against concurrent-request scenarios and Redis lock race conditions.
- Verified that the lock is query-specific and has a TTL to prevent deadlocks.
- Verified that lock ownership is checked before releasing the lock.
- Verified that the existing Overpass mirror-fetching behavior remains unchanged.
- Verified that only the intended file was modified.

### Environment Limitation

> [!NOTE]
> Automated Jest tests could not be executed locally because the current environment does not have the required `node_modules`/Jest dependencies installed.
>
> Running the test command resulted in:
>
> `jest is not recognized as an internal or external command`
>
> Therefore, no automated test result is being claimed locally.

## 🏷️ PR Type

- [x] 🐛 `type: bug`
- [x] ✨ `type: feature`
- [ ] 📖 `type: docs`
- [ ] 🧪 `type: testing`
- [ ] 🔒 `type: security`
- [x] ⚡ `type: performance`
- [ ] 🎨 `type: design`
- [ ] ♻️ `type: refactor`
- [ ] 🛠️ `type: devops`
- [ ] ♿ `type: accessibility`

### Distributed Request Coalescing

Concurrent requests for the same Overpass query now use a shared Redis lock:

```text
Request A ──> Cache MISS ──> Acquire Lock ──> Upstream Fetch ──> Cache
Request B ──> Cache MISS ──> Wait ────────────────> Cache HIT
Request C ──> Cache MISS ──> Wait ────────────────> Cache HIT

### DIFF
--- a/apps/web/app/api/overpass/route.ts
+++ b/apps/web/app/api/overpass/route.ts
@@ -1,5 +1,5 @@
 import { NextRequest, NextResponse } from "next/server";
-import { createHash } from "crypto";
+import { createHash, randomUUID } from "crypto";
 import { rateLimit } from "@/lib/rateLimit";
 import { getClientIp } from "@/lib/getClientIp";
 import { redis } from "@/lib/redis";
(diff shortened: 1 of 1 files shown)


dipexplorer:
quality: exceptional: for highly robust distributed Redis locking mechanism with LUA-based safe release and fallback polling

PR #4344, opened 2026-08-14
feat(voice): fetch supported languages from backend

### LINKED ISSUE #4318: [FEATURE]  Drive the Voice page's language list from the backend instead of a hardcoded list
### Which area does this feature belong to?

Frontend — UI / UX (apps/web)

### Estimated Difficulty

🟡 Intermediate — Some experience needed

### What problem does this solve?

GET /api/medicine/languages in apps/api/src/routes/medicine.ts forwards to the ML service's /voice/languages endpoint and returns whatever Indian languages the voice pipeline currently supports — it's mounted and live. But apps/web/app/[locale]/voice/lib/languages.ts hardcodes a fixed array of 6 languages (VOICE_LANGUAGE_OPTIONS) instead of calling that endpoint, and no file in apps/web calls /api/medicine/languages at all. This means if the ML service adds or temporarily drops a language, the Voice page's language picker has no way to reflect that — it's stuck at build time instead of reflecting what the backend can actually process.

### Proposed Solution

On the Voice page, fetch GET /api/medicine/languages on load and use it to filter VOICE_LANGUAGE_OPTIONS down to the languages currently supported by the ML service (matching on language/locale code), falling back to the full hardcoded list if the request fails or the service is unreachable — so the picker degrades gracefully rather than breaking. VOICE_LANGUAGE_OPTIONS stays as the source of frontend display metadata (labels, speech-recognition codes); the backend call only determines which of those options are currently selectable.

### Acceptance Criteria

_No response_

### GSSoC 2026

- [x] I am a GSSoC 2026 participant and would like to implement this feature.
- [ ] This feature uses **Cloudinary** (eligible for GSSoC Cloudinary bounty 🏆).

### Code of Conduct

OmanshiRaj:
This fixes the Voice page language picker being disconnected from the languages currently supported by the backend voice pipeline.

The Voice page now fetches the supported languages from `GET /api/medicine/languages` and uses the response to filter `VOICE_LANGUAGE_OPTIONS` based on the language/locale code. The existing `VOICE_LANGUAGE_OPTIONS` remains responsible for frontend display metadata such as language labels and speech-recognition codes.

If the backend request fails, times out, or the ML service is unavailable, the Voice page gracefully falls back to the complete hardcoded language list so that the language picker continues to work without interruption.

This ensures that when the ML service adds or removes support for a language, the Voice page can reflect those changes dynamically instead of relying only on a build-time hardcoded list.

Closes #4318


dipexplorer:
@OmanshiRaj, It appears that `/web/app/[locale]/page.tsx` (landing page) was completely overwritten and deleted, replaced entirely by your Voice component code. This breaks the application's homepage. Plz ensure u are only modifying the relevant component files without overwriting the core application pages, plz submit a fresh, clean PR.

